PR #9, opened 2022-06-11
adding confidence interval arguments

nabigac:
adding confidence interval arguments to the get_ks_stat, get_return_value, get_return_prob, and get_return_period functions! should be a pretty quick and easy pr to review but let me know if you have any questions!


vicford (commented):
Looks good to me!

threshold_tools_example.ipynb
```diff
@@ -3556,7 +3556,11 @@
     }
    ],
    "source": [
-    "return_value = threshold_tools.get_return_value(ams, return_period=10, distr='gev', multiple_points=True)\n",
+    "return_value = threshold_tools.get_return_value(ams, return_period=10, distr='gev', \n",
+    "                                                bootstrap_runs=100, \n",
+    "                                                conf_int_lower_bound=2.5, \n",
+    "                                                conf_int_upper_bound=97.5, \n",
```
thenaomig, line 3562: This is more of a comment for the threshold_tools function itself: rather than require all of these additional arguments in these function calls, which makes the whole thing seem more complicated to the notebook-user, would be better to have defaults on many of these arguments so that you don't *need* to specify them in the function call.

nabigac: i added notes explaining the default arguments for the confidence intervals and removed them for return probability and return period functions! let me know if it looks good or if there are any other comments!

PR #10, opened 2022-06-13
threshold tools application notebook for june working group

nabigac:
hi all, 

created and added the notebook that i'll be walking through for the threshold tools application. let me know if you see any changes and improvements that could be made and thanks for reviewing!


thenaomig:
A few things:
- if you choose ssp 2-4.5 or ssp 5-8.5 you will end up with only one GCM, so you don't need to select just one simulation (or scenario) by name. you can just .squeeze() the DataArray to remove singleton dimensions
- just noting that we do already have a card about unit conversion becoming a feature so you don't need to do it manually like this
- you can set it to '˚C' with the unicode ˚
- I don't think we need quite so many printouts at the end of cells for intermediate calculated quantities. That's something you'd just do while you were testing this. Ideally more of it moves to behind-the-scenes eventually, including getting the annual max series, which is an intermediate step.
- when you move to the statewide version of this same thing, you don't need to initiate another new app = ck.Application. That's something we envision happening once at the top of each notebook. 
- make sure all of these specific choices have been tested in the official jhub in the current version of main. I was getting an error (which is fixed if PR#45 is ever merged), so that it did not work for subsetting Sacramento in the current version of main in the official jhub

nabigac:
@thenaomig -- made the changes requested! 

still reviewing these two (if you have any comments/updates that would be great). thanks again! 

- if you choose ssp 2-4.5 or ssp 5-8.5 you will end up with only one GCM, so you don't need to select just one simulation (or scenario) by name. you can just .squeeze() the DataArray to remove singleton dimensions
- make sure all of these specific choices have been tested in the official jhub in the current version of main. I was getting an error (which is fixed if PR#45 is ever merged), so that it did not work for subsetting Sacramento in the current version of main in the official jhub

vicford (commented):
Minor things: 
- The order of selections suggested for app.select() in the threshold basics and the second threshold advanced sections can read better if the options are in a 'top-to-bottom' list as the user would select them; so: timescale, variable and spatial res, time slice, ending with the spatial selections. 
-  There's a typo in "topics" in the intro markdown code block

nabigac:
@vicford -- made the changes requested! 

also made some other updates to the notebook to further streamline. i think this should be good to go but let me know if there's any additional comments and if testing goes smoothly! @vicford @thenaomig

PR #11, opened 2022-08-08
Remove pip install line from top of notebooks

nicolejkeeney:
This PR just removes the pip install line from the top of the notebooks, following Brian's update (climakitae is now automatically loaded into the notebook environment in the AE engine). This will clean up the notebooks and avoid a lengthy install line. 

The +/- (+1878, -122) are quite large due to weirdness with jupyter notebooks; since I deleted an entire cell, it includes cell metadata in the +/-s. Hopefully that makes sense.

I'm unsure if the lmoments3 is still a dependency of the threshold_tools_basics notebook, so I've left that pip install in.

### DIFF
--- a/threshold_tools_application_examples.ipynb
+++ b/threshold_tools_application_examples.ipynb
@@ -55,11 +55,6 @@
    },
    "outputs": [],
    "source": [
-    "!pip install -v --upgrade git+https://github.com/cal-adapt/climakitae.git\n",
-    "\n",
-    "#install needed package to calculate parameters for extreme distributions via l-moments\n",
-    "#climakitae includes lmoments 1.0.1, which references scipy functionality which has been depreciated\n",
-    "#command below installs lmoments 1.0.4 which does not make references to missing/depreciated values\n",
     "!pip install git+https://github.com/OpenHydrology/lmoments3.git"
    ]
   },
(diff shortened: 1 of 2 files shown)


bkg:
Yeah, this is a little tricky to easily review since the diff is too large to display on GitHub, and I'm not sure this necessarily needs one.

As a hopefully helpful tip, using `git add --patch whateverfile.py` will allow staging significant lines instead of the entire file with possibly unimportant changes. This makes it simpler to review relevant changes, especially for `.ipynb` source which isn't really human readable. So feel free to try another PR with that if you like, or go ahead and merge this.

threshold_tools_application_examples.ipynb
```diff
@@ -1168,7 +1163,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.7"
+   "version": "3.8.3"
```
nicolejkeeney, line 1166: @bkg Any thoughts on what's going on here with the python version change? I just deleted some cells, not sure why this is occurring

bkg: It records the Python version used which is 3.9.7 on the pangeo notebook image, but it sounds like you probably have 3.8.3 locally.

nicolejkeeney: bleh. I think I'll try making another PR. Thanks for your help!

PR #13, opened 2022-08-11
Delete pip install line, clear cell outputs from notebooks

nicolejkeeney:
I just deleted the pip install line from each notebook. I also cleared the cell outputs from the threshold_tools_basics and timeseries_examples notebooks. I assume these were committed by accident in the past. The notebooks are now clean and don't show the cell outputs. They also now both render in GitHub (at least for me!). Let me know what you think.

### DIFF
--- a/getting_started.ipynb
+++ b/getting_started.ipynb
@@ -32,19 +32,6 @@
   {
    "cell_type": "code",
    "execution_count": null,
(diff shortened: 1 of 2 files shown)


vicford (approved):
Works on the hub

ckingdon95 (approved):
looks good, thanks for deleting the cell outputs!! been meaning to suggest that :)

PR #14, opened 2022-08-30
A new notebook for the 3rd working group, Aug 31 2022.

thenaomig:
A new notebook to showcase the feature in climakitae https://github.com/cal-adapt/climakitae/pull/65 for exploring warming levels. It also contains an application wherein we select a day-of-year and time-of-day with especially large changes, and plot that on a map with power station locations overlaid.

Reminder: when we merge this, it will appear in everyone's file list on the hub.


thenaomig:
I can't keep track of which working group we're on... but that's okay.

PR #18, opened 2022-10-14
Notebook update: Timeseries Tools 

vicford:
Splitting PRs per notebook, so apologies in advance for the multiple review requests. 

This PR updates our timeseries notebook for export and cleaned up some markdown text size. 

**To test**: 
1. Run through notebook
2. Check that export works for the timeseries data as a:
-  **netcdf**
-  **csv**
 
 GeoTIFF will not work as it is a timeseries, and the markdown text is updated to reflect this.

### DIFF
--- a/timeseries_example.ipynb
+++ b/timeseries_example.ipynb
@@ -42,6 +42,28 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "id": "54b4b6bf-8949-4155-ac6d-e9a686347278",
+   "metadata": {},
+   "source": [
+    "Additionally, get set up to make the computing go faster, by executing this cell:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "80537bea-0a3a-44ac-a5c8-4bef62f22e0c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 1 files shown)


gdicecco (approved):
I'm not sure how long this is taking for other people, but 3 & 9 km here for ~50 years of data w/ no spatial subsetting was taking more than 20 minutes to compute (that's when I quit kernel). This shouldn't preclude these changes from merging to main, but I think we eventually want a dask cluster chunk here too unless that wouldn't help with these particular computations.

Otherwise, no issues w/ select, explore, export!

thenaomig:
Yes, agree that we should also add a chunk of code for dask gateway to this
particular notebook.

vicford:
Added in a gateway cluster for this notebook - thanks for the suggestion Grace & Naomi! 

Will point out that working with daily data (20 years at 9km scale, for a county selection) will crash the kernel for me on testing this with no memory. Will revisit.

bethmints (changes requested):
just update to that text I sent you - sorry I couldn't push the PR myself, I think I have to generate a new key pair

elehmer (approved):
LGTM! Runs well and don't see any issues on the HUB with cluster.

PR #19, opened 2022-10-14
Notebook update: Threshold Basics

vicford:
Splitting PRs per notebook, so apologies in advance for the multiple review requests.

This PR updates our threshold basics notebook for export and removed an unnecessary import.

To test:

1. Run through notebook
2. Check that export works for the thresholds data as a:
- **netcdf**
- **csv**
- **GeoTIFF** will work as long as there is some prior subsetting (example is provided of this, but if there is a clearer example that could be substituted here).

### DIFF
--- a/threshold_tools_basics.ipynb
+++ b/threshold_tools_basics.ipynb
@@ -65,6 +65,28 @@
   },
   {
    "cell_type": "markdown",
+   "id": "0444b879-9e3c-4ba2-a830-57e36c88a30b",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


gdicecco (dismissed):
No errors with selections, transform code chunks, or exports. Also a good candidate for a dask cluster chunk if that works with the way the bootstrapping is set up behind the scenes

gdicecco (approved):
Sorry I got so mixed up between which notebook is which - this one works but could use dask if that works with the bootstrapping!

bethmints (approved):
markdown looks good, thanks!

PR #23, opened 2022-11-17
Replace .compute with app.load

nicolejkeeney:
This PR just replaces .compute() with app.load() in all notebooks. app.load() is a function I wrote that tells you how much data you're downloading before it is actually loaded into memory. 

explore_warming shows a large diff because the import statement in the previous version had an error raised in the output cell. I just cleared the kernel and pushed a fresh version. 

**To test:** Unfortunately, testing this PR will require running all the affected notebooks up to the app.load() call at the very least.


vicford (approved):
Ran into no issues in any of the five notebooks with app.load step, or the steps immediately after it

Tested locally, with a limited range of app.select options (when appropriate)

However, just noticing in threshold_tools_basics markdown `If you select 'daily' for 'Timescale' it will result in a daily average of the hourly data.` -- this is no longer true, yes?

PR #25, opened 2022-11-24
Fix wording to reflect removal of append historical button

nicolejkeeney:
Changed wording in affected notebooks. This PR is related to the changes made to app.select in climakitae [PR #119](https://github.com/cal-adapt/climakitae/pull/119) that removes Append Historical as a selection option. **This PR should not be merged until PR #119 is merged to climakitae**

### DIFF
--- a/timeseries_example.ipynb
+++ b/timeseries_example.ipynb
@@ -79,7 +79,7 @@
    "metadata": {},
    "source": [
     "## Step 1: Select\n",
-    "Be sure to select \"Append historical\" and \"Area average\" in order to work with the timeseries tools.\n",
+    "In order to work with the timeseries tools, **be sure to select both \"Historical Climate\" and one of the SSP timeseries options.** The historical data will be automatically appended to a SSP time series when both are selected. **You'll also need to select \"Area Average\".**\n",
     "\n",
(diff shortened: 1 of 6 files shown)


timeseries_example.ipynb
```diff
@@ -79,7 +79,7 @@
    "metadata": {},
    "source": [
     "## Step 1: Select\n",
-    "Be sure to select \"Append historical\" and \"Area average\" in order to work with the timeseries tools.\n",
+    "In order to work with the timeseries tools, **be sure to select both \"Historical Climate\" and one of the SSP timeseries options.** The historical data will be automatically appended to a SSP time series when both are selected. **You'll also need to select \"Area Average\".**\n",
```
vicford, line 82: More of just a comment of food for thought here: the timeseries notebook at present is designed to analyze the historical + SSP period together. The language here reinforces this (as it should). However, are we expecting none of our users will ever want to analyze just the historical period? 

Might be a question for @gdicecco?

vicford (approved):
Looks good to me!

timeseries_example.ipynb
gdicecco: I think this makes sense right now given how the tools are designed - if we get feedback that people may want to do historical only analyses we could revisit

gdicecco (approved):
Looks good to me!

PR #26, opened 2022-11-28
Remove markdown mention of TMY

gdicecco:
Outdated reference to when TMY was in panel in this notebook - just removed that sentence in the application section.


vicford (approved):
Good catch!

PR #27, opened 2022-12-13
Limit users to a single cluster

bkg:
Switch from using the `dask_gateway.GatewayCluster` to the climakitae `Cluster` which limits users to a single running cluster.


elehmer (approved):
LGTM! Working well on the HUB.

PR #30, opened 2022-12-16
Update AMY notebook with DFO-meteo_yr changes

vicford:
Updates the markdown and the relevant cells to update changes in the climakitae code in https://github.com/cal-adapt/climakitae/pull/139. 

Converting to a draft as new updates to the AMY notebook for the severe meteorological year functionality depends on on-going updates to climakitae. Will update once that PR is in.


elehmer (approved):

explore_amy.ipynb
```diff
@@ -18,12 +18,11 @@
...
     "2. Difference AMY: represents the difference between a designated future 30-year period and the historical baseline (1981-2010)\n",
-    "3. *Coming soon!* Severe meteorological year: represents the difference between a designated extreme percentile for a designated future 30-year period and the historical baseline (1981-2010)\n",
+    "3. Severe meteorological year: represents the difference between the 90th percentile for a designated future 30-year period and the historical baseline (1981-2010)\n",
     "\n",
-    "For any given location, we can calculate the desired AMY option and display one of the following three options depending on your selections:\n",
+    "For any given location, we can calculate the desired AMY option and display one of the following options depending on your selections:\n",
     "1. Heatmap of AMY\n",
-    "2. Diurnal cycle of a typical day within the AMY\n",
-    "3. Annual cycle of a typical hour within the AMY"
+    "2. A typical month of choice within the AMY"
```
nicolejkeeney, line 25: I think it might be nice instead to clarify here what the panel GUI can do, vs. what the notebook shows you how to do in code. For example, "Using the handy-dandy app.explore.amy() panel, you can easily set how you want to compute the average meteorological year and view a heatmap of the resulting average meteorological year. Later, we'll break down the calculation and show you how to code it up yourself to output a typical month of choice within an average meteorological year". Something like that.

vicford: I like this! I'll make the switch now

explore_amy.ipynb
```diff
@@ -182,7 +182,7 @@
    "source": [
     "my_data = app.retrieve() \n",
     "my_data = my_data.isel(simulation=0, scenario=0) \n",
-    "display(my_data) "
+    "display(my_data)"
```
nicolejkeeney, line 185: In the markdown right above (not sure how to highlight that portion), it says "Read the data into memory". Can you change that to something like "Read the raw hourly data into memory", then add into the description below "This is the raw data we will use to compute AMY in the next step." I just think it's a little confusing how it is now, with the user having two variables: `my_data` and `amy_data`, so I think it's good to emphasize the difference between the two data objects.

vicford: Done!

nicolejkeeney (approved):
looks good to me

PR #36, opened 2023-02-15
Minor update for warming_levels figure for new pangeo image

vicford:
Small updates for the warming levels notebook: 
1. Explicitly sets x and y, otherwise it barks
2. Sets the colormap to ae_orange
3. Updates language from axes to axs so scatter points can be overlain (this is the biggest difference)

**Setting as a draft for now until we officially switch pangeo images**


nicolejkeeney (approved):
Tested with the new pangeo image and works swimmingly!

PR #37, opened 2023-02-16
DFU feb16

nicolejkeeney:
Notebooks and data for DFU meeting Feb 16, including updates to hourly model notebook to include bias correction and replacing app.retrieve_from_csv with app.retrieve in all notebooks.


gdicecco (approved):
Tested the new hourly_model_inputs notebook on the hub and it works great!

PR #38, opened 2023-03-02
Removed xclim import from top of these notebooks

thenaomig:
for compatibility with the new hub image


nicolejkeeney (approved):
looks good to me

PR #41, opened 2023-03-11
Station 8760s

thenaomig:
Adds the notebooks: 'station_8760s.ipynb', which does two things:

- Access and export hourly data at station locations (netCDF or csv)
- Examine the “8760’s” and find the one with the smallest distance from the median -- view this in several ways.

The latter demo's another way to think about a "typical" 8760, as for an hourly load forecast.

### DIFF
--- a/specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
+++ b/specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
@@ -243,6 +243,8 @@
     "    to_return = y.sel(time=str(which_8760.year.values),simulation=which_8760.simulation.values)\n",
     "    # set time to the same generic calendar year regardless:\n",
     "    to_return['time'] = np.arange(1,8761)\n",
+    "    # drop the simulation coord \n",
+    "    to_return = to_return.drop(\"simulation\")\n",
     "    return to_return\n"
    ]
   },
@@ -283,7 +285,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "median_year.drop(['scenario','simulation']).hvplot()"
+    "median_year.drop(['scenario']).hvplot()"
    ]
   },
   {
@@ -401,7 +403,7 @@
     "median_year_array['hour'] = median_year_array['time.hour']\n",
     "median_year_array['dayofyear'] = median_year_array['time.dayofyear']\n",
     "\n",
-    "median_year_df = median_year_array.to_pandas().drop(['scenario','simulation'],axis=1) \n",
+    "median_year_df = median_year_array.to_pandas().drop(['scenario'],axis=1) \n",
     "median_year_df = median_year_df.pivot(index=\"dayofyear\", columns=\"hour\")\n",
     "median_year_df.columns.names = ['station', 'hour']"
    ]
--- a/specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
+++ b/specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
@@ -73,11 +73,11 @@
    "metadata": {},
    "source": [
     "#### Now retrieve the data, storing it in a variable called 'data'.\n",
-    "The data that will be retrieved is being bias-corrected using historical observations at each station selected. \n",
+    "The data that will be retrieved is being bias-corrected using hourly historical observations at each station selected. \n",
     "\n",
     "The historical observations come from [HadISD](https://catalogue.ceda.ac.uk/uuid/f579035b3c954475922e4b13705a7669). \n",
     "\n",
-    "The bias-correction procedure happening behind-the-scenes is called [QDM](https://journals.ametsoc.org/view/journals/clim/28/17/jcli-d-14-00754.1.xml), which is designed to preserve trends while allowing for different biases across quantiles. \n",
+    "The bias-correction procedure happening behind-the-scenes is called Quantile Delta Mapping ([QDM](https://journals.ametsoc.org/view/journals/clim/28/17/jcli-d-14-00754.1.xml)), which is designed to preserve trends while allowing for different biases across quantiles. \n",
     "\n",
     "The notebook 'localization.ipynb' steps through the bias-correction in more detail if you wish to understand what is happening."
    ]
@@ -141,11 +141,15 @@
    "id": "47d1abf5-1991-4fd9-8344-af5d7925e3ab",
    "metadata": {},
    "source": [
-    "We could stop there, but let's do a little more. Now that we have hourly data to work with, we can *improve upon the constructed 8760's that had otherwise been in use!*\n",
+    "Now that we have hourly data to work with, we can create and sample from a pool of 8760's which have consistent and realistic time series properties, improving on stochastic sampling methods that had otherwise been in use.\n",
     "\n",
     "#### Sample the \"8760\" closest to the median\n",
     "\n",
-    "One way that you may wish to use this data is to examine \"8760's\", or entire years of hourly data. If you chose a wide range of years previously, you might want to now sub-select, for example, only a historical reference period."
+    "One way that you may wish to use this data is to examine \"8760's\", or entire years of hourly data. If you chose a wide range of years previously, you might want to now sub-select, for example, only a historical reference period.\n",
+    "\n",
+    "Two main reasons to choose only a historical period (for now in this notebook):\n",
+    "- the following steps do not correct for the trends in the data due to climate change, so the median will reflect when the median amount of climate change has happened, instead of a median year solely with respect to year-to-year variability\n",
+    "- if you max out on the number of years, subsequent steps may execute more slowly"
    ]
   },
   {
@@ -218,7 +222,7 @@
     "\n",
     "def get_difference_hour(y): \n",
     "    #for this hour, of this day of the year, the median across all 8760s:\n",
-    "    median = y.median()\n",
+    "    median = y.quantile(q=0.5)\n",
     "    #return the difference from that median for each 8760:\n",
     "    return np.abs(y - median)\n",
     "\n",
@@ -247,7 +251,9 @@
    "id": "89537493-42ba-450c-bece-3dce51be677c",
    "metadata": {},
    "source": [
-    "And apply it across all the stations in the dataset:"
+    "Note that you could replace the 0.5 quantile (or median) with some other quantile.\n",
+    "\n",
+    "Apply the above functions across all the stations in the dataset:"
    ]
   },
   {
@@ -302,7 +308,7 @@
     "import pandas as pd\n",
     "\n",
     "def median_by_hour(y):\n",
-    "    by_hour = y.groupby('time.hour').median('time')\n",
+    "    by_hour = y.groupby('time.hour').quantile(q=0.5)\n",
     "    generic_year_day = pd.to_datetime('1981'+str(y['time.month'].values[0]).zfill(2)+\n",
     "                                      str(y['time.day'].values[0]).zfill(2))\n",
     "    by_hour['time'] = pd.date_range(generic_year_day,periods=24,freq='H')\n",
@@ -459,7 +465,8 @@
     "    y='hour',\n",
     "    yticks=[(int(one_hour),to_PST(one_hour)) for one_hour in median_year_array.hour.values],\n",
     "    groupby='station',\n",
-    "    cmap='Reds'\n",
+    "    cmap='Reds',\n",
+    "    clabel=app.selections.variable +' ('+ app.selections.units +')',\n",
     ")"
    ]
   },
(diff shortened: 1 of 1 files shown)


specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
```diff
@@ -0,0 +1,627 @@
...
+    "app.select()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7cb2fe31-d427-4fc4-8260-6b2e7b784e81",
+   "metadata": {},
+   "source": [
+    "#### Now retrieve the data, storing it in a variable called 'data'.\n",
+    "The data that will be retrieved is being bias-corrected using historical observations at each station selected. \n",
```
gdicecco, line 76: for clarity - suggest  to add '_hourly_ historical observations'

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
```diff
@@ -0,0 +1,627 @@
...
+    "\n",
+    "The historical observations come from [HadISD](https://catalogue.ceda.ac.uk/uuid/f579035b3c954475922e4b13705a7669). \n",
+    "\n",
+    "The bias-correction procedure happening behind-the-scenes is called [QDM](https://journals.ametsoc.org/view/journals/clim/28/17/jcli-d-14-00754.1.xml), which is designed to preserve trends while allowing for different biases across quantiles. \n",
```
gdicecco, line 80: can we write out quantile delta mapping first as well?

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
```diff
@@ -0,0 +1,627 @@
...
+   "source": [
+    "data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "47d1abf5-1991-4fd9-8344-af5d7925e3ab",
+   "metadata": {},
+   "source": [
+    "We could stop there, but let's do a little more. Now that we have hourly data to work with, we can *improve upon the constructed 8760's that had otherwise been in use!*\n",
```
gdicecco, line 144: Suggest something like: 
"Now that we have hourly data to work with, we can create and sample from a pool of 8760's which have consistent and realistic time series properties, improving on stochastic sampling methods that had otherwise been in use.

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
```diff
@@ -0,0 +1,627 @@
...
+   "id": "03bfdd3b-9fa3-46fe-b134-0ba4361ee826",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import xarray as xr\n",
+    "\n",
+    "def get_difference_hour(y): \n",
+    "    #for this hour, of this day of the year, the median across all 8760s:\n",
+    "    median = y.median()\n",
```
gdicecco, line 221: Unsure how easy this is to change out (and/or if this is a direction we want to go in at all), but it would be nice if this was a quantile function (and we could still show an example w/ 0.5 as the quantile), so the method was flexible and could pull e.g. a hot year for some percentile threshold

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
```diff
@@ -0,0 +1,627 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e2845243-040c-4c83-ac5d-e8f70c8991aa",
+   "metadata": {},
+   "source": [
+    "### Access\n",
+    "#### Start an application and launch the select panel\n",
+    "For the rest of the examples in this notebook, you will want to select 'station' for 'Data type'. Make any other choices you like."
```
gdicecco, line 56: I tried this for the greater bay area demand forecast zone, and in that zone Oakland Metro and San Francisco international airport stations seemingly had no data (returned all nan's) - not sure if this is a bug or desired behavior (if desired, we may eventually want to implement a warning message or something explaining what's going on for users)

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
```diff
@@ -0,0 +1,627 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f24d4452-8952-44cc-b5be-e3814c6bf261",
+   "metadata": {},
+   "source": [
+    "Now finally plot it all together. (Note that this plot is interactive and you can zoom in and out.) Initial plotting (and each time you switch station) may take a moment.\n",
+    "\n",
+    "Samples from all 8760's are plotted in light grey, the statistical medians for each of the 8760 hours of the year are in black, and the single 8760 with the smallest euclidean distance from that statistical median is shown on top in the thin line."
```
gdicecco, line 359: This plot hangs a lot for me when I try to switch stations and my browser tab goes unresponsive (to be fair, I did pick 1980-2050 originally so I have a lot of years in my data object). Maybe we should flag that for the user in the text

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
```diff
@@ -0,0 +1,627 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "65ea17fd-bba4-4a31-95a0-014065b4b7f3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "median_year_array.hvplot(\n",
```
gdicecco, line 457: A nice to have - label the color ramp with the variable & units represented

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
```diff
@@ -0,0 +1,627 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d9e1c694-2c40-4db7-9fba-8219306ae21a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data_table = data_table.unstack().unstack().stack('time')"
```
gdicecco, line 546: This structure is not ideal for reading into R for a few reasons - 
1) the nested header columns can't be easily represented in R data frame format (preference is for the structure shown in the cell above, L525, but with the scenario and simulation column values repeated for each row they apply to - this is the more standard way to work with csvs/'tidy' data in R which I know the DFU folks do)
2) spaces in column headers is a real pain in R as well - preference for snake or camel case if possible!

gdicecco (commented):
Made a few comments - in my view the highest priority would be in the CSV export step to make the data object more R friendly. There's also a few quick markdown suggestions, and some lower priority things I noticed while testing. 

Overall I think this will go over really well and meets what we know of their data needs!

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
thenaomig: That is not an expected behavior...

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
thenaomig: Yes, I've added a warning to stick with the historical period, both for speed and because this notebook is not accounting for trend in the data.

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
thenaomig: Not at all hard to replace .median() with .quantile(q=0.5), which may make it clearer how to modify this. (Also added a note that one could modify that.)

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
thenaomig: Done, although it's not done in the ideal manner in which I'd do it in a climakitae function, since I'm checking app.selections, and that could change in the meanwhile if people don't execute cells in order.

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
thenaomig: Please confirm that this is better.

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
thenaomig: Appears that this is an underlying bug with the data or climakitae code, which has been noted, and should not stand in the way of merging this notebook to have stakeholders test it.

vicford (approved):
No questions on my end, I especially like and appreciate having provided code to illustrate how to read the data files back in. I tested on a handful of stations and did not come across the missing data stations.

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
gdicecco: Hmm this version still reads out with some nesting in the column headers. I think the version we should recommend them to export is the one earlier in the notebook, in the chunk that reads:

`data_table = data.to_dataframe(dim_order=['scenario','simulation','time'])`

I checked that one in R, and although the preview of the data table in Jupyter does not repeat the scenario & simulation row values (which I was concerned about), the exported CSV does. So, I'd suggest just adding one more line to remove spaces in the column names (this is what I would do but maybe there's another way that's preferred):

`data_table.columns = data_table.columns.str.replace(' ', '_')`

And then doing the export to csv step for that file. The size is not an issue for reading into R, but we can keep the note about it being too large to open in Excel.

vicford:
Getting the following error at this cell `median_year = data.squeeze().map(get_median_8760)` specifically when testing Oakland and SF stations, which I had not gotten previously for other stations

`MergeError: conflicting values for variable 'simulation' on objects to be combined. You can skip this check by specifying compat='override'.`

specific_use_cases/DFU_work_in_progress/station_8760s.ipynb
thenaomig: Okay good to know. Will update as suggested.

gdicecco (approved):

nicolejkeeney:
> Getting the following error at this cell `median_year = data.squeeze().map(get_median_8760)` specifically when testing Oakland and SF stations, which I had not gotten previously for other stations. Note, I did see data for these two stations, not nans.
> 
> `MergeError: conflicting values for variable 'simulation' on objects to be combined. You can skip this check by specifying compat='override'.`
## Explanation of the issue
@vicford Nice catch. This is a weird little bug. I think this is unrelated to the previous issue of those stations returning nans. I think this is due to the fact that both stations reach the median year at different simulations. The function ``.map(get_median_8760)`` then tries to concatenate along the simulation dimension, but it can't, because they have different simulation values. The station subset worked for me actually if I chose a different temporal slice. Below I'll give an example of when it works, and when it doesn't, and then a proposed solution to address the problem. 

## Example of the problem
Info about the code snippets: 
- Both examples were run in the `bay_stations` branch of climakitae, see [this](https://github.com/cal-adapt/climakitae/pull/188) PR for more info. 
- Both examples were run for the subset defined by the Bay Area Demand Forecast Zone and the three stations contained within

### When it breaks
For the default temporal slice for historical data, we the stations reach the median year at different simulations. You can see that in the for loop. This will cause an error, which Victoria found. 
ERROR_different_simulations

## When it works
For a different time slice (I can't remember which), this same dataset of stations reaches the median year in the same simulation. The function works fine this way. Note how the simulation coord is the same--fgoals. 
WORKS_temporal_slice_1

## A solution 
Fixing this is easy (I think/hope?). Just modify the function ``get_median_8760`` to drop the simulation coordinate. 

```
def get_median_8760(y):
    difference_from_median = y.squeeze().groupby('time.dayofyear').apply(get_difference) 
    euclidean_distance = difference_from_median.groupby('time.year').apply(get_distance)
    final = euclidean_distance.stack(all8760s=['year','simulation'])
    which_8760 = final.isel(all8760s=final.argmin())
    to_return = y.sel(time=str(which_8760.year.values),simulation=which_8760.simulation.values)
    # set time to the same generic calendar year regardless:
    to_return['time'] = np.arange(1,8761)
    # drop the simulation coord 
    to_return = to_return.drop("simulation")
    return to_return
```
We end up dropping the simulation coord in the subsequent cell in the notebook (and at several other cells that follow that), so I don't think this will impact anything down the line.

vicford:
@nicolejkeeney - your fix works just fine for me. Logically, also because the simulation dim is dropped in the next cell I also don't see any issues with it being dropped one cell earlier to prevent this. I went ahead and made all the necessary changes for the rest of the notebook too, worth someone double checking that work before this gets merged in

nicolejkeeney:
> @nicolejkeeney - your fix works just fine for me. Logically, also because the simulation dim is dropped in the next cell I also don't see any issues with it being dropped one cell earlier to prevent this. I went ahead and made all the necessary changes for the rest of the notebook too, worth someone double checking that work before this gets merged in

Awesome, thanks for making the code changes!! I just ran the notebook in the hub and it works swimmingly. Feel free to merge!

PR #42, opened 2023-03-21
Notebook update for heating and cooling degree hours

vicford:
This updates the notebook associated with computing CDD+HDD, as well as providing new examples for cooling and heating degree hours. A new helper plotting function was also added to the .py file as well.

Please test this notebook with the `cddhdd_update` branch of climakitae: https://github.com/cal-adapt/climakitae/pull/191


specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -567,17 +591,27 @@
...
    "source": [
     "hdd_annual = compute_annual_aggreggate(\n",
-    "    data=hdd, \n",
+    "    data=hdd_aggregated, \n",
     "    name=\"Annual Heating Degree Days (HDD)\", \n",
     "    num_grid_cells=num_grid_cells\n",
     ")\n",
     "cdd_annual = compute_annual_aggreggate(\n",
-    "    data=cdd, \n",
+    "    data=cdd_aggregated, \n",
```
gdicecco, line 599: I think something is going wrong in how cdd_annual interacts w/ compute_multimodel_stats - the plot in 4e shows a multimodel min that's lower than any of the individual model outputs

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -690,12 +724,191 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1864932c-edd1-4ab1-9e4d-57b3ac74ffe8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hdh_cdh_lineplot(hdh_aggregated, start_date, end_date)"
```
gdicecco, line 847: This function is not defined - I don't see it in https://github.com/cal-adapt/climakitae/pull/191 either

gdicecco (changes requested):
A couple things to address before merging to main:

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
vicford: @gdicecco this function should be in utils_notebook_1.py script, not climakitae - sorry, should clarify that this py file will probably need to be updated on the hub as well! (copy and paste should be fine, I imagine)

gdicecco: I'm really unsure what the issue is actually - I updated utils_notebook_1.py, and can import the other functions there by name, but the hdh_cdh_lineplot one is erroring for me (`cannot import name`)? Have you tried this on the hub as well w/o issue?

vicford: I'll take a closer look into this!

vicford: @gdicecco - I checked on this on the hub and everything does work as desired for hdh_cdh_lineplot function. Will you confirm for me that the hdh_cdh_lineplot is in the utils_notebook_1.py again for me? Otherwise, I might suggest we try and see what is going on via Zoom

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
vicford: @gdicecco I have an answer! This is in fact the minimum, because FGOALS is uniquely across the board the minimum. If you toggle off the minimum line in the figure, FGOALS is directly underneath it. This also explains why the trend line is shifted seemingly downwards when looking at the other 3 models too. I also confirmed this by looking at cdd_aggregated too, and FGOALS is consistently the lowest value so this does track

Worth putting 1-2 lines of markdown to clarify this, or is that opening a can of worms about disregarding some models? 
Screen Shot 2023-03-23 at 10 02 47 AM
Screen Shot 2023-03-23 at 10 02 56 AM
Screen Shot 2023-03-23 at 10 09 14 AM

gdicecco: @thenaomig this is the issue I was referring to in standup - do we want to just alter the visualization a bit so its clear the fgoals and multimodel min lines are completely overlapping? Address in markdown? This plot I think will raise questions about fgoals in general so I'm unsure how we want to proceed.

thenaomig: Ideally we'd do this with bias-corrected data. (At the same time, it is concerning that it's so very biased with respect to the other simulations. We should discuss leaving it out of the catalog in future.) For now, I'd add something to the markdown about the fact that the gridded data is not currently bias-corrected and that as a result the min or max could just reflect a simulation that is biased high or low compared to the others.

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
thenaomig: I did not have any issue.

thenaomig (commented):
Good stuff. Suggested that we take the opportunity to clean up a few other things that could be superseded by the newer select/retrieve functionality for stations.

thenaomig:
One more thing, in the utils_notebook_1.py file:

- Perhaps the following functions should be in climakitae in utils, just slightly more general: 'compute_multimodel_stats' and 'trendline'

- the repetition (and specificity to hdd and cdd) in 'compute_multimodel_stats' is unneccesary. You could be computing those across the simulation dimension for any dataarray. It should take one dataarray at a time (thus in this notebook it would be called twice in a row)

- in trendline, might speed it up if you can avoid the need to access .values ... is there a built-in line-fitting function for that next step?

Goal would be to do away with this file eventually

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
thenaomig, line 1: consider leaving out the cluster because 1. it's pretty fast without it, and 2. if you take my suggestion to use the new station functionality in app.select then it won't currently work with a cluster anyway

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -690,12 +724,191 @@
...
+   "id": "12f6395e-ee86-46a9-b631-e2495c7af283",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# ALL DATA WITHIN DFZ ZONE\n",
+    "data_to_use = data_dfz\n",
+    "num_grid_cells = data_dfz.x.size * data_dfz.y.size # Number of grid cells within the demand forecast region"
+   ]
+  },
+  {
```
nicolejkeeney, line 760: Neither `data_to_use` nor `num_grid_cells` is used after this point, so I recommend removing this cell as to not confuse people.

specific_use_cases/DFU_work_in_progress/utils_notebook_1.py
```diff
@@ -67,4 +67,13 @@ def hdd_cdd_lineplot(annual_data, trendline, title="title"):
         color="black", 
         line_dash='dashed', 
         label="trendline"
-    ) 
\ No newline at end of file
+    ) 
+
+def hdh_cdh_lineplot(data, date_start, date_end):
+    data_to_plot = data.loc[:, date_start:date_end]
```
nicolejkeeney, line 73: My preference would be for the plotting function to just perform the plotting, and have the time slice occur directly in the notebook. I also really like xarray's selection functionality that allows you to just use "June 2011" to get the same data you have here-- it's super easy to read/use and might empower users to play with it more? Just my two cents!

```
cdd_one_month = cdh_aggregated.sel(time="June 2011")
hdh_cdh_lineplot(cdd_one_month)
```

nicolejkeeney: Plus, I think the plotting function should work regardless of whether or not you want to plot a particular time slice

nicolejkeeney (approved):
Looks great, I left some suggestions regarding the time slicing and the grid cell thing.

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -523,16 +528,16 @@
...
-    "**HDD = 65 - temperature<br>\n",
-    "CDD = (-1)\\*(65 - temperature)**<br><br>\n",
-    "For HDD, we can just subtract the 2m temperature from 65 degrees Farenheight, then set any negative to 0. For CDD, we will do the same, but will then multiply by -1 to turn negative values to positive, then set negative values to 0. We need to multiply by -1 for CDD to avoid having all negative values; for example, on a day of 80F, CDD = 65 - 80 = -15, but the CDD value is +15. Multiplying -15 by -1 will give us the true value of 15. "
+    "We'll use the climakitae helper function `compute_hdd_cdd` to compute both heating and cooling degree days, which uses a default temperature of 65 degrees F for both metrics. You can modify the default options using the function arguments `hdd_threshold` and `cdd_threshold` to represent any threshold of your choosing. In the example below, we will calculate HDD with a threshold of 60 degF and CDD with a threshold of 70 degF. The function performs the following calculations:<br><br>\n",
```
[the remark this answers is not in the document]
vicford: The hdd/cdd and hdh/cdh functions now accept thresholds for different thresholds, but I've also removed any reference to 65 degF, except that it is the default value. Do we want to modify the function so that the user must pass a value instead and remove all 65 degF references?

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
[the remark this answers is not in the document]
vicford: @nicolejkeeney can you provide insight on the RH question here? Is it because that file is used elsewhere in other DFU notebooks?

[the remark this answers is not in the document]
nicolejkeeney: No, from my memory it's because it was asked of us by the demand forecast unit. But since we're not using anything except 2m temp in the notebook, i don't see any reason not to just use app.select() here and just pull temp. They could always re-run the notebook for whatever variables they want.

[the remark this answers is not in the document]
nicolejkeeney: I don't see why not. 

> *



> One more thing, in the utils_notebook_1.py file:
> 
> * Perhaps the following functions should be in climakitae in utils, just slightly more general: 'compute_multimodel_stats' and 'trendline'
> * the repetition (and specificity to hdd and cdd) in 'compute_multimodel_stats' is unneccesary. You could be computing those across the simulation dimension for any dataarray. It should take one dataarray at a time (thus in this notebook it would be called twice in a row)
> * in trendline, might speed it up if you can avoid the need to access .values ... is there a built-in line-fitting function for that next step?
> 
> Goal would be to do away with this file eventually

I also agree that we should remove this file eventually

gdicecco (approved):
Markdown changes look good to me!

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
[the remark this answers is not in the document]
thenaomig: Only remaining thing would be to remove the now unnecessary import: from climakitae.selectors import Boundaries

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -129,253 +92,192 @@
...
   },
   {
    "cell_type": "markdown",
-   "id": "a16bc16c-d454-47f8-81d7-1f360e6c46df",
+   "id": "a68132d7-e91c-4555-ab0d-13af7d61e1b7",
    "metadata": {},
    "source": [
-    "Next, we need to convert the lat/lon coordinate pair to the model's projection coordinates. We can easily do this using the built in helper function in climakitae: `get_closest_gridcell`. For more information on this function, you can call `help(get_closest_gridcell)` or look in the climakitae.utils module for the actual code that performs the computation."
+    "### 1c) Output final data product as a csv file\n",
+    "We'll drop all unneeded coordinates and convert our xarray Dataset to a pandas Dataframe, allowing us to easily output the final data product to a csv file. In the output table, the first column is the time in units of UTC, and the second column are the various global climate models (which can be filtered in excel or in python code in the notebook). The other columns are the variables selected at the beginning of the notebook."
```
thenaomig, line 164: There are some extra unneeded coordinates that do not appear (any longer) that it can't find to drop. And the 'station_name' is not recognized for replacing ' ' with '_'

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -129,253 +92,192 @@
...
+    "In this example, we will visuzalize the data across the Demand Forecast Zone for the Sacramento Municipal Utility District, and then calculate the median of all conditions across the Sacramento Municipal Utility District."
    ]
   },
   {
    "cell_type": "markdown",
    "id": "0bffd2d1-104a-4dbd-a308-1ed4efdee246",
    "metadata": {},
    "source": [
-    "### 3c) Visualize both the Demand Forecast Zone and the weather station on the same map "
+    "### 3a) Visualize both the Demand Forecast Zone and the weather station on the same map "
```
thenaomig, line 272: Doesn't recognize 'one_station' for this location plot -- you could put back something reading that in, but just specify Sacramento instead of searching for the station in the zone. Alternatively you could direct the user to view this in app.select, although they wouldn't get to appreciate the data grid-spacing that way.

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
thenaomig: (similar thing happens in 3c also)

thenaomig: (and 4g)

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -129,253 +92,192 @@
...
-  },
-  {
-   "cell_type": "markdown",
-   "id": "1d4cd9b4-6df8-499a-b7d9-c45c782b3b87",
-   "metadata": {},
-   "source": [
-    "### 2b) Grab the closest grid cell to the weather station\n",
-    "To demonstrate this process, we'll use the Sacramento Executive Airport weather station."
+    "%%time\n",
+    "data_closest_gridcell = app.load(data_closest_gridcell)"
```
thenaomig, line 145: let's call this data_at_station or something like that instead now

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -623,7 +555,9 @@
    "metadata": {},
    "source": [
     "### 4f) Visualize the results\n",
-    "Using the python package *hvplot*, we can easily make a line plot of the annual aggregated data. To do this, we'll plot the annual HDD, then add the trendline on top. The code to generate the plot is contained in a function `hdd_cdd_lineplot`. "
+    "Using the python package *hvplot*, we can easily make a line plot of the annual aggregated data. To do this, we'll plot the annual HDD, then add the trendline on top. The code to generate the plot is contained in a function `hdd_cdd_lineplot`. \n",
```
thenaomig, line 558: hdd lineplot just looks like a flat line... 
hdd_lineplot

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -692,21 +626,142 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5c55c615-93e0-4710-8501-bf168979e33d",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "### 5b) Display a month of CDH and HDH\n",
```
thenaomig, line 686: I'm seeing all the plots in this section as flat lines.

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
vicford: @thenaomig I can't reproduce this and do get the plots to run for me

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
thenaomig: Whatever changed (could be my library version) it looks good now.

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
thenaomig: Okay, it works for me now too.

PR #44, opened 2023-04-04
Reverts Warming Level notebook for older pangeo image

vicford:
Please test that this notebook runs with this image in https://github.com/cal-adapt/climakitae/pull/196


elehmer (approved):
This works on the reverted image

PR #45, opened 2023-04-17
Modify to match simulation naming in loca update

nicolejkeeney:
The explore_internal_variability notebook uses assumptions about the names of the simulations that are no longer correct with the new loca branch in climakitae, which adds the downscaling method to the simulation name (i.e. "CESM2" becomes "CESM2_WRF"). This PR fixes that and slightly cleans up the related code chunk. 

**To test:** Make sure you run the notebook in the climakitae branch `loca`


vicford (approved):
No issues on my end!

bethem (commented):
worked for me!

PR #47, opened 2023-04-19
Modify simulation selection to match with LOCA catalog

nicolejkeeney:
Needed to modify how the simulations are subsetted again since the naming of the simulations changed using app.retrieve() for the new data catalog structure. 

Test using climakitae branch `loca_ensemble`


vicford (approved):
Looks good to me

PR #53, opened 2023-05-19
Improve outdated markdown instructions and warming levels bug

nicolejkeeney:
The thresholds_basics notebook has instructions that reference an old setup of app.select, which had two panels. I just slightly modified the wording to fix this.

Also, this PR fixes a broken plotting situation in explore_warming. I think the URL to the powerplant shapefile had been changed (we should probably not link to external sites for this reason!) 

I meant to do these in two separate branches but my Friday brain got mixed up, which is why they are in the same PR.


gdicecco (approved):
Looks good, thanks Nicole!

PR #59, opened 2023-06-20
Add new notebook for custom extreme event types (threshold_tools)

corakingdon:
This pull request:
- creates new notebook: "threshold_tools_event_types.ipynb", needs to be tested using the [`thresholds` branch](https://github.com/cal-adapt/climakitae/tree/thresholds) of `climakitae`
- renames the calls to `get_ams()` to `get_block_maxima()` in existing thresholds notebooks to match the updated thresholds functionality from [this climakitae PR](https://github.com/cal-adapt/climakitae/pull/240)

Seeking any/all feedback, especially on guiding text, especially on the discussion around effective sample size!


work_in_progress/threshold_tools_basics.ipynb
vicford, line 1: Finding at this step that I have to app.load the data first before running threshold_tools.get_block_maxima otherwise it throws this error

>  ams = threshold_tools.get_block_maxima(subsetted_data, extremes_type='max')
>  ams = app.load(ams) 

`ValueError: apply_ufunc encountered a dask array on an argument, but handling for dask arrays has not been enabled. Either set the ``dask`` argument or load your data into memory first with ``.load()`` or ``.compute()```

threshold_tools_event_types.ipynb
vicford, line 1: For selections in app.select, it says select 3-7.0... no historical as well? Since it's hourly is there a specific year's or year range worth of data we are looking to test with?

vicford: Seeing now that it should be for a much longer timespan, like 2020-2100. I think we should include language to that effect just so its not ambiguous!

vicford, line 1: Getting `'bms2_3d' is not defined` and `bms2_4h3d is not defined` errors in part 3, looks like the 2 is not in either of those original set-ups. Also noticing there's a mix of "ams" and "bms", is that intentional?

vicford, line 1: Seeing that the return period for example 2 is nearly 4000 years, and that the last two examples have nans provided. We should include further markdown explaining why that may be the case, and potentially some final concluding thoughts as to how to use these numbers in workflows? The two explore_uncertainty notebooks have good example of "take away points of how to use these tools" that could be used as a baseline

threshold_tools_event_types.ipynb
```diff
@@ -0,0 +1,505 @@
...
+    "ams = get_block_maxima(hourly_data_subset)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ams.plot()"
```
gdicecco, line 181: This plot needs more explanation - ideally write in markdown that its plotting the hottest hour per year

gdicecco: Also, for me the scenario got cut off in the title

threshold_tools_event_types.ipynb
```diff
@@ -0,0 +1,505 @@
...
+    "    groupby = (1, 'day'), \n",
+    "    grouped_duration = (3, 'day')\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Note: effective sample size\n",
```
gdicecco, line 238: Ah this is really helpful! I would suggest just moving this up in the notebook to when the user might first encounter the warning message

threshold_tools_event_types.ipynb
```diff
@@ -0,0 +1,505 @@
...
+    "**Return value for example 3, hottest temperature reached 3-days in a row:**"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "get_return_value(bms2_3d, return_period=20, multiple_points=False)"
```
gdicecco, line 387: `bms2_3d` is not defined for me

threshold_tools_event_types.ipynb
```diff
@@ -0,0 +1,505 @@
...
+    "**Return value for example 4, hottest temperature reached for 4 hours each day 3 days in a row:**"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "get_return_value(bms2_4h3d, return_period=20, multiple_points=False)"
```
gdicecco, line 403: `bms2_4h3d` not defined either

threshold_tools_event_types.ipynb
```diff
@@ -0,0 +1,505 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Return periods\n",
```
gdicecco, line 410: It would be nice to show at least one visualization for the return value and return period objects

gdicecco (changes requested):

threshold_tools_event_types.ipynb
corakingdon: I'm wondering if you selected data in a different unit than Fahrenheit? the value of 105 being provided is somewhat arbitrary, but needs to correspond to Fahrenheit data!

corakingdon: I just renamed the misnamed variables, sorry! the `ams` was meant to refer to the two examples that kept a block size of 1 year (so it's still an annual series) then I used `bms` for the examples that used a different block size... but definitely open to thoughts on if that should be changed or clarified!

work_in_progress/threshold_tools_basics.ipynb
corakingdon: interesting, I'm not exactly sure what this error means, but when I saw these notebooks I was surprised that the `load` comes after `get_block_maxima` anyway, so I am somewhat inclined to just move the load step earlier in the notebook on the raw subsetted data (to make all of the following calculations faster). thoughts?

threshold_tools_event_types.ipynb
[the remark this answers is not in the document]
corakingdon: sounds good, will work on this!

corakingdon: yes good point, will also add more specific text to the selection part!

threshold_tools_event_types.ipynb
```diff
@@ -0,0 +1,505 @@
...
+    "\n",
+    "The following four examples show how to use these options to construct different types of events of interest. See section 3 for further analysis and discussion of these events."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Example 1: hottest hour in each year\n",
+    "The basic use case of `get_block_maxima` pulls the maximum value in each year of data, shown here:"
```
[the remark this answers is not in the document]
corakingdon: so the estimated ~23 is the average ESS per year (or per block if block_size is different than 1 year). So it's not about the number of years, but the effective sample size within each year. This has to do with the amount of autocorrelation and the type of event (if you define a 10-day long event, the sample size in each year is smaller than an hour long event). 

I hadn't run into a case of getting an average ESS <25 for example 1 though. Do you remember what your other selections were? this was hourly data right?

threshold_tools_event_types.ipynb
vicford: I can't replicate it now, so probably just user error on my part!

threshold_tools_event_types.ipynb
corakingdon: sorry! just updated

threshold_tools_event_types.ipynb
corakingdon: will fix this!

threshold_tools_event_types.ipynb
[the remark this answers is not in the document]
gdicecco: Hourly, temperature in F, 1980-2100 for SSP 3-7.0 for a county w/ area average for Amador county

Using average ESS across all years is an interesting diagnostic - temporal autocorrelation of temperature at least is not necessarily constant through end of century (depending on how you're calculating it), so I'm curious whether there is any directional trend in ESS across years that might bias that average?

threshold_tools_event_types.ipynb
vicford: Ah, okay, that makes sense! Maybe just for clarification point, I think a simple inclusion to the markdown would help in case others have questions here (or does it not matter?)

Maybe "we'll call our annual series "ams" as it has a block size of 1 year, and where we modify the block to be a different block size we'll call it "bms" " for transparency.

nicolejkeeney (changes requested):
Please add a link to this notebook and a short description of it to the AE_navigation_guide.ipnyb notebook under the appropriate section. Otherwise looks good! :D

threshold_tools_event_types.ipynb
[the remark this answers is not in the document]
corakingdon: I played around with this, and you're totally right-- depending which county you select, and whether or not you include historical climate data in the selected data, you can end up with the "too small ESS" warning in the first two examples, as you found. I think this is tricky because we don't know when the user will first encounter the error, so for now, I'm inclined to make "Sacramento County" the official focus of this example notebook so that the explanatory text for ESS stays at the 3rd example, where the warning will first be thrown. Also, depending on which area you use, the example return value of 105F (in the Step 3 EVA section) may not be an informative value to use (might end up with very large return periods), but it is a good value to use for Sacramento. I added a note addressing this, and added a comment in the selection section to use Sacramento county for the initial exploration in this notebook.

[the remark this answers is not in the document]
gdicecco: I think that's a reasonable plan! It makes sense to not want to bog down the reader w/ all the details right away

gdicecco (approved):

vicford (approved):
Noting that some of the code in the notebook will change with the upcoming refactor, but as is, looks great (I can assist/update once the refactor is in)

PR #64, opened 2023-08-18
Fix for the INTRO notebook for calculating HDD/CDD with station data

vicford:
The INTRO_annual_consumption model previously provided the option to calculate HDDs/CDDs using either an aggregation over the DFU zone, or to use the closest gridcell to the weather station itself, however much of the notebook was configured for the gridded data and would break if station gridcell option was selected instead. 

**Note 1**: Our previous calculation for CDD/HDD was incorrect as it did not resample the hourly to daily data to calculate CDD/HDD accurately for either gridded or station data, this would produce reasonable but low values in the gridded data, and several orders of magnitude too large in the station data. This is now fixed by resampling. 

**Note 2**: The plot of the gridded station data with the overlay station scatterpoint **does not work** - I believe this is because app.view now returns a panel object instead. This needs to be fixed.

---
To test:
Run the INTRO notebook for both options in Step 4 (i.e., test once with the default behavior in Step 4a, and then the "closest grid cell" option) and make sure data is produced. This notebook also needs to be run in the `fix_dd` climakitae branch - also in PR here: https://github.com/cal-adapt/climakitae/pull/273

### DIFF
--- a/specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
+++ b/specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
@@ -454,14 +454,14 @@
    },
    "outputs": [],
    "source": [
-    "# # ALL DATA WITHIN DFZ ZONE\n",
-    "# data_to_use = data_dfz.resample(time='1D').mean() # Resamples to daily data\n",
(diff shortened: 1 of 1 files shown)


jmlongmate (approved):
Notebook runs for me as well with both gridded and station data, order of magnitude and trends make sense to me-- agree with Grace that the order of magnitude looks about right! Nice.

PR #66, opened 2023-08-29
Update CDD/HDD Notebook for daily mean method options, update refactored code, markdown

vicford:
Per request, have modified the daily mean calculation to include options: daily mean, and the max-min average, with a source link for the max-min method. Also updated a few lingering refactored code snippets. 

To test:
- Can either run the notebook, or just look at the code changes, they're minor!


specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -427,7 +423,7 @@
...
...data within the demand forecast zone, which we retrieved in **step 2b**. Or, you can use the closest grid cell to the weather station, which we computed in **step 1b**. We do not recommend using the aggregated DFZ data calculated in step 3b, as aggregating the data prior to computing HDD and CDD may remove some critical information about the weather extremes. You can comment out whichever method you don't want to use. We've chosen to show the analysis with the DFZ data, but if you want to use the closest grid cell data, just comment out the DFZ cells and uncomment the closest grid cells. Note, the closest grid cell resampling will take a few minutes! <br><br>Depending on the input data, we will also set a new variable defining the number of grid cells. This will be just 1 for the closest grid cell method; for the DFZ data, however, this value will change depending on the size of the DFZ before aggregating. This information is used to compute the annual aggregate HDD and CDD in step 4c. Lastly, we provide 2 options for determining the [daily mean](https://www.weather.gov/key/climate_heat_coolhttps://www.weather.gov/key/climate_heat_cool) for which we calculate degree days against. "
```
jmlongmate, line 426: rephrase "for which...against" to "against which..." (does that preserve the meaning well?)

jmlongmate, line 426: link pasted twice, redirects

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -440,15 +436,32 @@
...
+   "source": [
+    "# METHOD 1: Using daily max/min difference\n",
+    "daily_max = data_to_use.resample(time='1D').max()\n",
+    "daily_min = data_to_use.resample(time='1D').min()\n",
+    "data_to_use = (daily_max + daily_min)/2\n",
+    "\n",
+    "## METHOD 2: Using daily mean\n",
+    "# data_to_use = data_to_use.resample(time='1D').mean()"
+   ]
+  },
```
jmlongmate, line 464: looks good!

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
jmlongmate: Also in the weather.gov page linked here, I found this sentence useful for describing why we calculate HDD/CDD: "Degree days are based on the assumption that when the outside temperature is 65°F, we don't need heating or cooling to be comfortable." 

This notebook seems to be pretty specific and assumes that the user knows why they want to calculate HDD/CDD, and wants that threshold temperature to be flexible-- is it worth including something like this in the text introducing Step 4 as a reminder of why the user might set a certain temperature threshold ?

jmlongmate (commented):
Looks great! Mostly just need to fix the link to the daily mean calculation, feel free to take or leave my comment about describing rationale for setting a temperature threshold from which to calculate HDD/CDD.

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
vicford: Great catch, fixed! And yes, I think that would be super useful context to include in the markdown!

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
```diff
@@ -417,7 +413,7 @@
...
    "source": [
     "## Step 4: Compute heating degree days and cooling degree days\n",
-    "Here, a heating degree day (HDD) is calculated by computing how many degrees Fahrenheit **colder** the daily temperature is from a specified temperature threshold. A cooling degree day (CDD) is calculated by computing how many degrees **warmer** the daily temperature is from a specified temperature threshold. In the computation below, you can provide different thresholds for HDD and CDD based on your needs. "
+    "Degree days are [based on the assumption](https://www.weather.gov/key/climate_heat_cool) that when the outside temperature is 65°F, we don't need heating or cooling to be comfortable. However, you may wish to use a different threshold than 65°F. In the code below, a heating degree day (HDD) is calculated by computing how many degrees Fahrenheit **colder** the daily temperature is from a specified temperature threshold. A cooling degree day (CDD) is calculated by computing how many degrees **warmer** the daily temperature is from a specified temperature threshold. In the computation below, you can provide different thresholds for HDD and CDD based on your needs. "
```
gdicecco, line 416: Just to further prime readers to expect the function to allow two different thresholds, maybe make this say: "However, you may wish to use a different threshold than 65°F - for example, you might want to assume that temperatures between 60-70°F won't require heating or cooling, and use degrees below 60°F for HDD< and degrees above 70°F for HDD)." 

You could put a line break before "In the code..." if you think the paragraph is getting too long w/ this change

gdicecco (approved):
One minor markdown suggestion to further clarify our fn allows two different thresholds for HDD and CDD, otherwise looks good to me!

specific_use_cases/DFU_work_in_progress/INTRO_annual_consumption_model.ipynb
vicford: I like this addition @gdicecco!

PR #70, opened 2023-09-12
Updated warming level notebook

thenaomig:
To use with the warming_upgrade branch of climakitae.

I also made some of the markdown text more concise. There was a lot of repetition with what it also says in the panel.

### DIFF
--- a/explore_warming.ipynb
+++ b/explore_warming.ipynb
@@ -277,7 +277,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "my_gwl = 3.0  # 3˚ global warming"
+    "my_gwl = 2  # 2˚ global warming"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


vicford (commented):
Looks good to me. A couple of thoughts:
- We might want to include some markdown on the gwl_snapshots and sliced_data pieces in the notebook as to how these can be used --> can be a future iteration of this notebook though
- Should maybe rename "test" in `test = wl.sliced_data` to something else, maybe "my_wl_data" ?

explore_warming.ipynb
```diff
@@ -85,9 +84,129 @@
...
+   "source": [
+    "wl.visualize()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "183deba2-9aca-45ee-bed8-8ae90c4db735",
+   "metadata": {},
+   "source": [
+    "wl.visualize() step, above, is optional if you want to go directly to extracting the data."
```
claalmve, line 139: nitpick (not critical to change but could be nice): Re-word to "The `wl.visualize()` step above is optional if you want go directly to extracting the data."

Tianchi-Liu (commented):
In the Export section, the `wl.retrieve()` could use an update to another output the users might want to export. 

Unfortunately, xarray cannot export `to_plot` as it, because of the error `variable 'all_sims' is a MultiIndex, which cannot yet be serialized to netCDF files`. To export it, a restructure is needed.

vicford:
On part 2 application:
- For the sake of not having empty plots in the part 2 application, we should set the warming level to 2° instead of 3° as the default option which will produce 6 plots when using dynamical data
- Part 2 at present is not formatted for use with the LOCA2 data if the design is to pull hourly data, should include something in the markdown to that effect

thenaomig:
> On part 2 application:
> 
>     * For the sake of not having empty plots in the part 2 application, we should set the warming level to 2° instead of 3° as the default option which will produce 6 plots when using dynamical data
> 
>     * Part 2 at present is not formatted for use with the LOCA2 data if the design is to pull hourly data, should include something in the markdown to that effect

Changes made.

PR #74, opened 2023-10-11
Heat Index Notebook

vicford:
This notebook walks through how to obtain and interpret heat index values. 

---
**To test**
Run through the notebook and let me know scientific + programmatic thougths!

### DIFF
--- a/collaborative/IOU/heat_index.ipynb
+++ b/collaborative/IOU/heat_index.ipynb
@@ -336,7 +336,7 @@
    "source": [
     "### Step 2: Calculate the number of days each year above a Heat Index threshold\n",
     "\n",
-    "As we noted above, the Heat Index only \"kicks in\" once the Heat Index value is above 80°F: prolonged exposure to a heat index above 80°F becomes dangerous to many people and especially vulnerable communities. We'll now calculate the number of days in each year (i.e., the summer period only) that are above a specific threshold. Because there are specific thresholds used in the Heat Index [classification](https://www.noaa.gov/sites/default/files/2022-05/heatindex_chart_rh.pdfhttps://www.noaa.gov/sites/default/files/2022-05/heatindex_chart_rh.pdf) system, we'll reproduce the NOAA guidance below and we **strongly recommend** looking at multiple thresholds to understand Heat Index trends. We will start with 80°F.\n",
+    "As we noted above, the Heat Index only \"kicks in\" once the Heat Index value is above 80°F: prolonged exposure to a heat index above 80°F becomes dangerous to many people and especially vulnerable communities. We'll now calculate the number of days in each year (i.e., the summer period only) that are above a specific threshold. Because there are specific thresholds used in the Heat Index [classification](https://www.noaa.gov/sites/default/files/2022-05/heatindex_chart_rh.pdf) system, we'll reproduce the NOAA guidance below and we **strongly recommend** looking at multiple thresholds to understand Heat Index trends. We will start with 80°F.\n",
     "\n",
     "| Classification | Heat Index |\n",
     "|----------------|------------|\n",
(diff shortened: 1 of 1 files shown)


claalmve (approved):

claalmve:
I think the only change/reminder was to add the markdown warning about the runtime for downloading hourly data, but otherwise looks good to me!

Tianchi-Liu:
Great to see heat index in action! I have one major concern about the non-local time that starts at about 4pm Pacific Time. While the counts of extreme "days" still capture the trends, their definition is less intuitive than if local hours are used. Plus the use of this non-local time would inflate the count of extreme days in step 2 (say a day with a few "very warm" hours that sit right across this non-local time zone's midnight, one extra "very warm" "day" might be counted compared to when the local time zone is used).

collaborative/IOU/heat_index.ipynb
```diff
@@ -0,0 +1,557 @@
...
+    "print('The projected change in the median Heat Index value from historical is: {:.2f}°F'.\n",
+    "      format(hi_proj - hi_hist))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2f4e4b46-5312-47e5-99fc-95fec3d828cc",
+   "metadata": {},
+   "source": [
+    "### Step 2: Calculate the number of days each year above a Heat Index threshold\n",
```
Tianchi-Liu, line 345: Nit: ## would suffice, same for Step 3

collaborative/IOU/heat_index.ipynb
```diff
@@ -0,0 +1,557 @@
...
+   "id": "09fba876-5f97-4262-b80e-a5ceb226e8db",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "selections.latitude = (stn_lat - 0.05, stn_lat + 0.05)\n",
+    "selections.longitude = (stn_lon - 0.05, stn_lon + 0.05)\n",
+    "\n",
+    "# because we're retrieving a single grid cell, we also take the area average\n",
```
Tianchi-Liu, line 150: could be multiple grid cells (e.g. Burbank-Glendale-Pasadena Airport (KBUR)), which is no problem. The values would just be weighted area averages.

Tianchi-Liu (commented):
The later comments on the notebook are minor compared to the time zone one. They don't have to be addressed as I understand the notebook is still under development!

vicford:
> I have one major concern about the non-local time that starts at about 4pm Pacific Time. While the counts of extreme "days" still capture the trends, their definition is less intuitive than if local hours are used. Plus the use of this non-local time would inflate the count of extreme days in step 2 (say a day with a few "very warm" hours that sit right across this non-local time zone's midnight, one extra "very warm" "day" might be counted compared to when the local time zone is used).

@Tianchi-Liu this is a really good point. Unfortunately, we don't have a clean way to do a local timezone conversion yet, and it's becoming increasingly more relevant for multiple notebooks -- I've got a trello card up for this issue. I have tested a "brute force" way but definitely need assistance on its improvement and implementation as a function we can call when needed. For now, I think as this notebook is still very preliminary and that the stakeholder in particular was most interested in the long-term trends over time between a projected future and the historical period, we would be fine for an approximation at present, and can work on refining the values for the next iterations of this notebook.

collaborative/IOU/heat_index.ipynb
Tianchi-Liu: get_closest_gridcell makes sense! I noticed the `selections.cached_area` is set to PG&E, so get_closest_gridcell wouldn't return the expected data if the input lat/lon is outside of the PG&E zone. If you think the stakeholders might do that for some reason, we'd need a workaround. 

This line below gives the PG&E stations you listed. Feel free to take or leave.
`wx_stns[wx_stns.station.str.contains('|'.join(['Arcata', 'Red Bluff', 'Stockton', 'San Jose', 'Fresno', 'San Luis Obisbpo', 'Bakersfield']))]`

Tianchi-Liu (approved):

PR #75, opened 2023-10-13
Latest GWL Notebook

claalmve:
Adding the latest version of the Global Warming Levels notebook to tools. This notebook's use is to demonstrate the changes between using the new Global Warming Levels framework vs. traditional SSP frameworks to understand global warming, through visualizations, statistical models, and bootstrapping. Visualizations also used in CPUC presentation.

**What to look out for:**
Ways to summarize the notebook, make the notebook more readable, or notes on changing details about visualizations.

**Notes on running notebook:**
Please run this notebook on the hub with Dask clusters! I have tried a couple times without and it has consistently crashed for me (lack of space, time-out).

Please also have the `gwl_upgrade` branch running as the climakitae back-end to this notebook.

### DIFF
--- a/tools/warming_levels_approach.ipynb
+++ b/tools/warming_levels_approach.ipynb
@@ -40,14 +40,6 @@
   },
   {
    "cell_type": "markdown",
-   "id": "9b89c286",
-   "metadata": {},
-   "source": [
-    "If you're interested only in finding warming data specific to your local area, please check out our `warming_levels.ipynb` notebook also in the tools folder of our `cae-notebooks` directory! You'll be able to find the file in the file directory on your left."
-   ]
-  },
-  {
-   "cell_type": "markdown",
    "id": "07c0eb87-5be0-4447-8c02-722dc6476412",
    "metadata": {
     "tags": []
@@ -122,7 +114,7 @@
    "id": "7539acd2-be4a-4b95-98ff-6cd35c75829f",
    "metadata": {},
    "source": [
-    "Now, let's launch a toolkit to view localized projections under varying levels of warming. First, choose a variable, then a spatial area of interest. "
+    "Now, let's launch a toolkit to view localized projections under varying levels of warming. We are going to pre-select specific selections that will effectively showcase the differences between the Traditional SSP approach and the New GWL approach."
    ]
   },
   {
@@ -148,14 +140,6 @@
    ]
   },
   {
-   "cell_type": "markdown",
-   "id": "745129d9-17f4-4b95-bc14-78719f8fdcfa",
-   "metadata": {},
-   "source": [
-    "A latitude and longitude of interest:"
-   ]
-  },
-  {
    "cell_type": "code",
    "execution_count": null,
    "id": "8069a8ba-9e13-4522-b745-46a3683b4c96",
@@ -164,49 +148,14 @@
    },
    "outputs": [],
    "source": [
-    "my_lat, my_lon = 34.08214634521255, -117.2425643"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "a3e5890c-a9f0-4dd2-9fd6-0bfa67a150fa",
-   "metadata": {},
-   "source": [
-    "A range that will result in the nearest gridcell being included:"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "2203f24d-4406-4b33-99e8-fbb65d5b0d5d",
-   "metadata": {
-    "tags": []
-   },
-   "outputs": [],
-   "source": [
-    "wl.wl_params.latitude=(34.0,34.3)\n",
-    "wl.wl_params.longitude=(-117.5,-117.1)"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "faa0ad08-88d4-447d-a34b-cb2035feda7d",
-   "metadata": {},
-   "source": [
-    "And let's set a few other things for this example, in case we forget to do so above.\n",
+    "# Selecting a specific latitude and longitude of interest:\n",
+    "my_lat, my_lon = 34.08214634521255, -117.2425643\n",
     "\n",
-    "If you'd like to keep your parameters from the panel above, **do not run the below cell,** as this will overwrite your custom inputted parameters from above."
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "32fa4079-74f1-42e0-93ee-8b47c0183d95",
-   "metadata": {
-    "tags": []
-   },
-   "outputs": [],
-   "source": [
+    "# Specifying a range that will result in the nearest gridcell being included:\n",
+    "wl.wl_params.latitude=(34.0,34.3)\n",
+    "wl.wl_params.longitude=(-117.5,-117.1)\n",
+    "\n",
+    "# Setting a few other parameters to extract warming levels:\n",
     "wl.wl_params.variable=\"Maximum air temperature at 2m\"\n",
     "wl.wl_params.units=\"degF\"\n",
     "wl.wl_params.timescale=\"daily\"\n",
@@ -219,7 +168,7 @@
    "id": "e126830a-c919-4af0-818f-2936daff2712",
    "metadata": {},
    "source": [
-    "Depending on your parameters, this calculate step <span style=\"color:red\">**may take up to ~45 min. to complete!**</span>. If you select statistically downscaled data on the highest resolution, this will take the longest, as it will deal with the largest amount of data and number of simulations. We are currently developing data optimization techniques to speed this up, but for now, when you run this cell, feel free to leave for a bit and come back in ~45 min. to check on how it's doing.\n",
+    "Depending on the parameters, this calculate step <span style=\"color:red\">**can take up to ~35 min. to complete!**</span>. If you were to select statistically downscaled data on the highest resolution (which is what we have selected above), this will take the longest, as it will deal with the largest amount of data and number of simulations. We are currently developing data optimization techniques to speed this up, but for now, when you run this cell, feel free to leave for a bit and come back in ~35 min. to check on how it's doing.\n",
     "\n",
     "In case you are curious, you can additionally open the Dask dashboard link above to view the computation that is being done, so you can visually tell if computation is currently happening."
    ]
@@ -237,7 +186,15 @@
   },
   {
    "cell_type": "markdown",
-   "id": "8ec9685c-939c-437b-bcf5-b0512e1a63f0",
+   "id": "ca76e369-5af3-4933-8429-e54fd070f0f1",
+   "metadata": {},
+   "source": [
+    "<span style=\"color:red\">**Disclaimer:**</span> We recognize that some of the following analyses, conclusions, and figures are dependent on the data parameters specified above to an extent. So, you may get different results and figures based on different data parameters you select above, but you should be able to arrive at the same conclusions we arrive at at the end of this notebook."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "4dbab622-480a-4798-bfab-f3dd5964832a",
    "metadata": {},
    "source": [
     "### (Optional): Visualize the regional response at a series of global warming levels."
@@ -260,17 +217,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "wl.visualize()"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "e925104f-57b3-40e6-8487-c9b42c555559",
-   "metadata": {},
-   "source": [
-    "- Step 1: Explore SSP framework\n",
-    "- Step 2: explore new framework\n",
-    "- Step 3: Compare the two different approahces"
+    "# wl.visualize()"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


claalmve:
There is a section in this notebook that discusses the statistical significance between two confidence intervals of two fitted distributions. The conclusion I had come away with (as of how it is now in the notebook) was that there was no statistical significance between the two intervals. However, I think there is a better way to convey the conclusion that a tighter confidence interval is better, but I'm not sure how to do so. So, I'm very open to suggestions on how to do so here!

vicford (commented):
Leaving some preliminary comments while waiting for the data to load in Step 2. 

My read on this notebook is that this notebook is a more in-depth explanatory notebook than the original warming_levels.ipynb by illustrating the differences between looking at the "traditional" SSP time-based approach for future periods of interest and comparing to this "new" GWL warming levels approach. 
- If this is the case, is this notebook intended to replace warming_levels.ipynb?
- There needs to be a very clear explanation of the intention of this notebook in the Introduction markdown. It wasn't immediately clear to me what this notebook was doing at first, as well as the language surrounding the "traditional" approach and the "new GWL" approach. At first, I was thinking it was the "traditional" warming levels approach and we were proposing an updated warming levels approach, which is why I have been confused.
    -  We might want to consider illustrating the SSP approach first, and then show WL as the improvement to have the notebook flow better. 
    - I had a lot of "why are we doing this exact step/code" questions when reading the rest through, that I think could have added markdown to clearly explain steps. It's a beefy and technical notebook, and many of our users will have questions along the way that we should assist with. I can help with this @claalmve!

Minor, but important:
- Naming of the notebook dependent upon whether this will replace the old WL notebook. If not, we should make sure this notebook name is very explicit about order of user use
   - Once this is confirmed, also add into the AE_navigation_guide
- Add an intended application to match our current structure
- The "steps" outline stops after Step 2, and I really think the entire notebook could benefit from this structure as it's a really hefty notebook!
- Before this gets merged, best practice is to remove the cell output so it is easier to track broad notebook changes through github

vicford (changes requested):
Two major things for this notebook: 
1. Markdown must be updated to explain the steps and outcomes of this notebook. A lot of the "why are we doing this next step" is missing. I can help with this, if you need @claalmve. 
2. All the sections in the "Examining Differences Between both approaches" do not currently tell a convincing story as to why the GWL approach is better than the time-based SSP approach. A lot of this is tied to the markdown itself

Going to leave a bunch of individual code-based comments for easier resolving/tracking

tools/updated_warming_levels.ipynb
vicford, line 1: There's an import for `from climakitae.explore.warming import fit_models_and_plots` twice

vicford, line 1: There's a typo in the visualize panel (so climakitae side), needing a space between "each" and "GCM"

vicford, line 1: We still need to come up with labels + colorbar for the wl.visualize postage stamps, as well as not providing all 100+ models in a single column

vicford, line 1: Why was `two_degrees` loaded (and other steps further in the notebook)? Wasn't the data already loaded with the `calculate()` step, or am I missing something?

vicford, line 1: For the centered year histogram plot, we can set up bins using the centered_year min and max to show the distribution a bit more clearly. Also don't need the plt.show() lines as they're rendered in the notebook.

vicford, line 1: Will point out if you take the median of the two_degrees centered_year values, that it is also 2040 which is used in the traditional approach comparison.

vicford, line 1: Is there a specific reason why 115°F was chosen for the exceedences? This is a fairly arbitrary selection, but that is a fairly high temperature and some justification for it's selection would be good to provide for users.

vicford, line 1: Can bypass finding the original time window piece by just using `wl.wl_params.window` in the next cell too

vicford, line 1: The counting # of excessive heating days per year figures are a bit confusing with the partial day x-axis, can we modify this to be integers? I think there is markdown here needed to explain the interpretation of this figure and then we can highlight the long tail

vicford, line 1: The y-axis on the comparison figure then needs to also be "number of years" not count

vicford, line 1: I think we should be really intentional and clear throughout this notebook about the usage of "new", "original", "traditional". 
All instances of time based "old" method should be stated as such: "Time-based SSP"
All instances of GWL method: "Warming Level"

vicford, line 1: Missing import for scipy.stats for ttest_ind

vicford, line 1: The t-test here for significance shows that the confidence intervals are not statistically significant at all, which is unsurprising. Fitting the logscale plot to all non-zero counts of excessive heat days hides what we are trying to demonstrate here in that the GWL method does in fact have a higher sampling frequency to detect extremes in the tails. Based on this figure, it shows that they're statistically identical which is not what we want to convey

vicford, line 1: In the boostrapping section, "heated year" is referring to what?

vicford, line 1: Unsure of what we're trying to convey with the boostrapping histogram + GWL probability. I believe the intention here is to show that GWL will catch very edge cases, and boostrapping will not

vicford, line 1: I would revise the last section of markdown about `clearly this probability is very significantly less than 0.05` and statistical significance as we're not assessing significance here at all and should not make scientific claims as such. The point here is that the boostrapped method has a <<<<0% probability of capturing an excessive heating day as compared to GWL which can.

vicford (changes requested):

tools/updated_warming_levels.ipynb
claalmve: This is a fantastic point. I forgot about this, yes, loading twice would be quite redundant.

claalmve: Like num_bins = (max - min) / 10 ? Or some number?

claalmve: Should I include that as a note in the centered year histogram?

claalmve: 😅 Yes you are very right. I was worried about this. I can remove this then.

claalmve: Yes! I can provide more context in the notebook to describe that.

claalmve: Referring to the likelihood of predicting a future year with > 10 excessive heating days. 10 was kind of arbitrarily chosen. I could re-do this metric to instead measure 10 consecutive days? To measure the likelihood of predicting future heatwaves? Or I could leave it as is and just describe what a "heated year" is more clearly.

vicford: More food for thought that they are identical, this is a science thinking Q!

vicford: Ah, we should include that in the markdown!

vicford: I was just confused as to what a "heated year" was, some description would be good!

claalmve:
I think `warming_levels.ipynb` is missing a section about how to extract this data for a specific lat/lon. I'm going to come back to that part later.

mdkoenig:
[Warming Levels Notebook Debugging and Cleaning up Code](https://trello.com/c/voZZERx2/494-warming-levels-notebook-debugging-and-cleaning-up-code)

Tianchi-Liu:
The default data of the `wl.choose_data()` UI breaks `wl.calculate()`. Maybe consider moving up the example data (that we know work well) parametrized in the cells below and/or suggesting to users to select those data in the UI. Also recognizing that the analysis, its conclusions, and figures are dependent on data selection to some extent, maybe warn users that results might vary and figures might need edits if different data are selected.

tools/warming_levels_approach.ipynb
```diff
@@ -0,0 +1,952 @@
...
+   "source": [
+    "new_params, trad_params = fit_models_and_plots(np.log1p(gwl_dist_vals), np.log1p(time_period_vals), 'pearson3');"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ac5d8ce1-b887-42f6-a7af-51ea8dcb91a8",
+   "metadata": {},
+   "source": [
+    "We can see that there is significant overlap between the two probability distributions of Extreme Heat Days counts, but it is also clear that there is a difference between the two distributions' confidence intervals."
```
Tianchi-Liu, line 760: The `fit_models_and_plots` function gives this figure:
Screenshot 2023-11-14 095430

I think It would be helpful to include one or two sentences on the interpretation of the 95% CI lines, as well as how the figure shows that GWL is more effective than the traditional method.

Tianchi-Liu:
Thoughts on Step 4.2: I am not sure the analysis here provides enough evidence of the conclusions. I like the considerations for the statistical significance, but I have a few concerns. 
1. In the final figure, the comparison between the two methods is not fair because the "New GWL probability" doesn't account for counts that are zeroes while "Bootstrap probability" does (if I read the code correctly). I would suggest making it consistent across 4.2, whether to include zero or not. 
2. Another aspect of the comparison that might be less than fair is that a distribution is generated for the bootstrap probability, seemingly to account for its uncertainties, but not for the new GWL probability. If that's indeed the intent, maybe consider the uncertainties in the new GWL probability as well.

Tianchi-Liu:
> Thoughts on Step 4.2: I am not sure the analysis here provides enough evidence of the conclusions. I like the considerations for the statistical significance, but I have a few concerns.
> 
> 1. In the final figure, the comparison between the two methods is not fair because the "New GWL probability" doesn't account for counts that are zeroes while "Bootstrap probability" does (if I read the code correctly). I would suggest making it consistent across 4.2, whether to include zero or not.
> 2. Another aspect of the comparison that might be less than fair is that a distribution is generated for the bootstrap probability, seemingly to account for its uncertainties, but not for the new GWL probability. If that's indeed the intent, maybe consider the uncertainties in the new GWL probability as well.

I think the last paragraph is an overstatement based on my reasonings above, but let me know if I'm missing something!

tools/warming_levels_approach.ipynb
```diff
@@ -0,0 +1,952 @@
...
+   "source": [
+    "## Step 2: Extract warming level data using new GWL method"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ca9653fb-e680-42f6-8bc6-9760405ac977",
+   "metadata": {},
+   "source": [
+    "Let's extract warming level data using the new GWL method. **We will once again select 30-year slices centered around 2040, or the year when SSP3-7.0 reaches 2˚C** above pre-industrial levels, according to the IPCC weighted census projection."
```
Tianchi-Liu, line 379: I am confused why the 30-year slices are centered around 2040. Don't different models have different center years?

tools/warming_levels_approach.ipynb
```diff
@@ -0,0 +1,952 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "931cd5c5-be63-49b9-9efb-948447984b70",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "# Finding the exceedance count for our data from the SSP approach.\n",
+    "time_period_dist = get_exceedance_count(ssp_data, threshold_value)\n",
```
Tianchi-Liu, line 469: `threshold_value` is not defined.

tools/warming_levels_approach.ipynb
```diff
@@ -0,0 +1,952 @@
...
+   "outputs": [],
+   "source": [
+    "# Plotting distribution of excessive heating by original SSP method\n",
+    "plt.figure(figsize=(6,3))\n",
+    "plt.hist(time_period_dist.values.flatten())\n",
+    "plt.title('Counting number of excessive heating days by year')\n",
+    "plt.xlabel('Number of excessive heating days')\n",
+    "plt.ylabel('Number of years')\n",
+    "\n",
+    "### NOTE: Make the days not partial days, only integers. Add Markdown to explain the interpretation of this figure and highlight the long tail."
```
Tianchi-Liu, line 498:
```suggestion

```

tools/warming_levels_approach.ipynb
```diff
@@ -0,0 +1,952 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "5302c374-5b5c-436e-8acd-874579546572",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "### Log-scaled plots\n",
```
Tianchi-Liu, line 667: Nit: x label is log(1 + count) to be exact

tools/warming_levels_approach.ipynb
```diff
@@ -0,0 +1,952 @@
...
+    "\n",
+    "# Show percentage of sampling a distribution with excessive heating days\n",
+    "heat_wave_days = 10\n",
+    "bootstrap_prob = len(bootstrapped[bootstrapped > heat_wave_days]) / len(bootstrapped)\n",
+    "new_prob = len(flat_new[flat_new > heat_wave_days]) / len(flat_new)\n",
+    "# bootstrap_prob = np.percentile(bootstrapped, 95)\n",
+    "# orig_prob = np.percentile(gwl_dist_vals.flatten(), 95)\n",
+    "\n",
+    "print(\"Bootstrapped probability of a future year with > 10 excessive heating days:\", bootstrap_prob)\n",
+    "print(\"Warming Level probability of a future year with > 10 excessive heating days:\", orig_prob)"
```
Tianchi-Liu, line 807:
```suggestion
    "print(\"Warming Level probability of a future year with > 10 excessive heating days:\", new_prob)"
```

tools/warming_levels_approach.ipynb
```diff
@@ -0,0 +1,952 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e6bfaa67-4add-4395-b0c7-2550adf0df99",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Now, we run bootstrap n=1000 times and calculate how many sampled distributions encapsulated extreme cases of heating > a given number of days.\n",
```
Tianchi-Liu, line 847:
```suggestion
    "# Now, we run bootstrap n=10000 times and calculate how many sampled distributions encapsulated extreme cases of heating > a given number of days.\n",
```

Tianchi-Liu (commented):
Love the addition of all the markdown texts! See more comments below.

claalmve:
> The default data of the `wl.choose_data()` UI breaks `wl.calculate()`. Maybe consider moving up the example data (that we know work well) parametrized in the cells below and/or suggesting to users to select those data in the UI. Also recognizing that the analysis, its conclusions, and figures are dependent on data selection to some extent, maybe warn users that results might vary and figures might need edits if different data are selected.

This is a good point. I think the scope of Warming Levels functionality is very limited right now, so it will be good to list to the user specific criteria that this notebook would work under.

tools/warming_levels_approach.ipynb
claalmve: Yes, this is true. I don't know why I wrote this. Re-writing now.

PR #80, opened 2023-11-30
Adds timezone correction to Heat Index and HDD/CDD notebooks

vicford:
Adds the `convert_to_local_time` correction to the heat_index and annual_consumption_model notebooks as they use hourly data that require daily min/maxes being on the "correct" timestamp for that day. 

----
**To Test**
Run these two notebooks, at least up to the timezone correction

### DIFF
--- a/collaborative/IOU/heat_index.ipynb
+++ b/collaborative/IOU/heat_index.ipynb
@@ -37,8 +37,10 @@
    "source": [
     "import climakitae as ck\n",
     "import pandas as pd\n",
+    "from climakitae.util.utils import read_csv_file, get_closest_gridcell, compute_multimodel_stats, trendline, summary_table, convert_to_local_time\n",
     "\n",
-    "from climakitae.util.utils import read_csv_file, get_closest_gridcell, compute_multimodel_stats, trendline, summary_table"
+    "import warnings\n",
+    "warnings.filterwarnings(\"ignore\")"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


collaborative/IOU/heat_index.ipynb
```diff
@@ -49,7 +51,7 @@
...
+    "First we'll grab specific locations of interest, by using the latitude and longitudeconvert_to_local_timeions of a weather station, and provide code to input a custom lat-lon location. Furthermore, we will **not** be retrieving the actual station data that is bias-corrected to that station for this example. At present bias-corrected station data on the Analytics Engine only provides air temperature as a variable, and for Heat Index we must also have either dew point temperature (coming soon!) or relative humidity. So for the time being, we will retrieve **non-bias corrected** data at the location of interest."
```
claalmve, line 54:
```suggestion
    "First we'll grab specific locations of interest, by using the latitude and longitude locations of a weather station, and provide code to input a custom lat-lon location. Furthermore, we will **not** be retrieving the actual station data that is bias-corrected to that station for this example. At present bias-corrected station data on the Analytics Engine only provides air temperature as a variable, and for Heat Index we must also have either dew point temperature (coming soon!) or relative humidity. So for the time being, we will retrieve **non-bias corrected** data at the location of interest."
```

claalmve: I'm not sure if you meant to include `convert_to_local_time` in this sentence differently, but I figured it's a typo as it is right now.

vicford: This is a good catch - definitely a typo, and some overzealous copy & paste

claalmve:
LGTM! Ran both notebooks on the hub and they worked with timezone conversions. Wooo!

PR #81, opened 2023-12-06
Adds simulation finder tool figure 1 notebook to WIP

vicford:
This is a trimmed down version of Julia's variance_calculations notebook. AE navigation guide is also updated to incorporate this.


AE_navigation_guide.ipynb
claalmve, line 1: Is there a reason for removing `threshold_event_types.ipynb` from the navigation guide?

work_in_progress/model_variance.ipynb
```diff
@@ -0,0 +1,691 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fa85b245-a1f8-4603-ac51-472445e34292",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "from climakitae.explore.uncertainty import weighted_temporal_mean\n",
```
claalmve, line 129: Could remove the commented code and move the `weighted_temporal_mean` function to `utils` now that multiple files reference it.

AE_navigation_guide.ipynb
vicford: This is a good catch - I'm not sure why it was deleted but adding back in

work_in_progress/model_variance.ipynb
vicford: It's already in utils, just forgot to remove this commented code out, good catch!

PR #97, opened 2024-04-02
Adding changes to agnostic_tools.ipynb to reflect changes in back-end for WRF V1.1

claalmve:
Notebook changes in agnostic tools to accommodate for V1.1 of visualizing WRF simulations. Making simulation ranking tool for agnostic tools more flexible to allow for additional types of WRF inputs.

This notebook can be tested with back-end changes found in this [PR](https://github.com/cal-adapt/climakitae/pull/334).


vicford (changes requested):
A few things:
- I would have the default option for this plot be temperature x precipitation, as that is a very common example of this kind of figure (i.e., temperature is the first selected metric, precipitation is the second)
- Both Step 2 runtime cells take less than a minute, so the markdown about this taking 10+ minutes can be removed, yay for efficiency!
- I would move cell 13 (where a second metric is selected) down into as the first cell of Step 4
   - There needs to be markdown explaining these two figures for the notebook and why a second metric is needed, given that this notebook is live on the hub - we need to provide documentation to our process

work_in_progress/agnostic_tools.ipynb
```diff
@@ -387,10 +402,10 @@
    "outputs": [],
    "source": [
     "# Plotting distribution of simulations based on if your method was 'WRF' or 'LOCA'.\n",
```
Tianchi-Liu, line 404:
```suggestion
    "# Plotting distribution of simulations based on if your downscaling method was 'Dynamical' or 'Statistical'.\n",
```

Tianchi-Liu (commented):
Nice additions of metric options and WRF visuals. Below are one major concern and a couple minor comments.

A major potential source of confusion is the difference between lat_lon and area_subset. Right now only `results_area` is plotted. Some guidance would be helpful to users of `results_gridcell`. Or if the notebook is ready to be generalized, consider combining `agg_lat_lon_sims` and `agg_area_subset_sims` and give the results the same name (ideally more descriptive than "results"). Similarly, Step 3 only showcases `single_stats_gridcell`.

Again, when this notebook is ready to be generalized, consider moving all parameters defined in Step 2 to Step 1.

Mentions of SCE Service Territory should be updated to California.

Tianchi-Liu:
> * Both Step 2 runtime cells take less than a minute, so the markdown about this taking 10+ minutes can be removed, yay for efficiency!

LOCA calculations still take long, but I agree that the WRF runs are fast, so the WRF users can be freed from the warning.

claalmve:
> > * Both Step 2 runtime cells take less than a minute, so the markdown about this taking 10+ minutes can be removed, yay for efficiency!
> 
> LOCA calculations still take long, but I agree that the WRF runs are fast, so the WRF users can be freed from the warning.

Added text to describe the expected run-time difference between running `LOCA` and `WRF` data.

work_in_progress/agnostic_tools.ipynb
```diff
@@ -269,30 +282,9 @@
...
-   "execution_count": null,
-   "id": "bb30abff-dfbd-4f99-8c20-63975a037b0e",
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "%%time\n",
-    "metric = \"Average Annual Total Precipitation\"\n",
-    "single_stats_area2, multiple_stats_area2, results_area2 = agg_area_subset_sims('states', 'CA', metric, years, method, months)"
+    "area_subset = 'states' # Choose your `area_subset`\n",
+    "selected_area = 'CA' # Choose your `selected_area`\n",
```
claalmve, line 286: At this point, in order for the users to know which areas are available to select on, they would need some version of the `Select` object available to them. I think for demo purposes, we should leave this as is, but I think this again strongly suggests that the `Select` object be used here in some detail in order to validate arguments and showcase options to the user.

Tianchi-Liu:
numpy is used in np.mean and needs to be imported.

`plot_double_WRF` is now `plot_climate_response_WRF`

vicford:
Is `multiple_state_gridcell` supposed to return no simulations in the default behavior?

Screenshot 2024-04-17 at 10 31 12 AM

vicford:
What are the defaults we want to display/discuss in this notebook? At present they are set for surface pressure and total precipitation. At least for WRF data

>I would have the default option for this plot be temperature x precipitation, as that is a very common example of this kind of figure (i.e., temperature is the first selected metric, precipitation is the second)

vicford:
Noting on the LOCA2 test run for Step 2 it took closer to 16 min, so we need to refine the warning

vicford:
Testing with statistical data to see the `plot_sims` and getting this error:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [36], line 5
      3     plot_WRF(results_area, variable, agg_func)
      4 elif downscaling_method == 'Statistical':
----> 5     plot_sims(results_area, variable, agg_func, years, single_stats_area)

TypeError: plot_sims() takes 4 positional arguments but 5 were given
```

claalmve:
> Is `multiple_state_gridcell` supposed to return no simulations in the default behavior?
> 
> Screenshot 2024-04-17 at 10 31 12 AM

Addressed! This is a specific issue of finding the middle 10% of simulations when there's less than 10% of simulations to choose from. I added more fallback logic to just return the median simulation if that case comes up again.

PR #98, opened 2024-04-03
Removing cluster cell from `warming_levels_approach.ipynb`

claalmve:
Removing cluster cell since the cluster is not working properly.


vicford (approved):
Great, thanks @claalmve

PR #103, opened 2024-05-01
Adding total run time to WL approach notebook and making small modifications

claalmve:
Adding text about overall runtimes for running a notebook from start to finish for warming levels notebooks. Additionally, this PR re-writes part of `warming_levels_approach.ipynb` to decrease the overall runtime for the whole notebook.


vicford:
We need to be careful about how LOCA2 data is represented in the visualizing step, especially if it's a single point location because...
Screenshot 2024-06-12 at 5 22 37 PM

This is in the warming_levels_approach NB with climakitae [PR #343](https://github.com/cal-adapt/climakitae/pull/343). When you zoom in (substantially) it improves to show a bar plot of all LOCA2 simulations
Screenshot 2024-06-12 at 5 26 36 PM

vicford (changes requested):
Okay, with default settings 
- cell 11 [the two two_degreees.sel(lat=my_lat, lon=my_lon, method='nearest')] is breaking because the two_degrees data doesn't have a lon dimension, only a double lat 
   - i.e., two_degrees is 21900 days_from_center x 3 lat x 129 all_sims
   - dropping the `squeeze` from the cell above fixes this problem, however...  
- then getting this error on cell 18
Screenshot 2024-06-12 at 5 40 56 PM

vicford:
Side note, the initial loading step takes closer to 45-50 minutes for me, not 35.

vicford:
So the notebook runs in entirity, BUT: 

Screenshot 2024-06-14 at 1 12 52 PM

claalmve:
> So the notebook runs in entirity, BUT this ends up being 120+ rows of this....
> 
> Screenshot 2024-06-14 at 1 12 52 PM

Yeah, these plots are tricky for LOCA. This is technically expected (2x2 gridcells being visualized), but of course it is not that interpretable. I will keep the notebook defaults to a single gridcell with LOCA for run-time purposes, but I will work on making sure that the above images visualize as a 4xn grid instead in case a user does select an area of gridcells in their parameters instead.

vicford (approved):

PR #104, opened 2024-05-01
Upgrading `agnostic_tools.ipynb`

claalmve:
The following changes should be observed in this PR:

1. 8 hourly WRF sims vs 4 monthly WRF sims observed in gridcell aggregations in the latter plots. This is mainly seen through the usage of the `wrf_timescale` variable being `monthly` vs `hourly`.
2. Climate response plot being rendered as an hvplot, allowing users to toggle specific GCMs or simulations, depending on which type of downscaling method is used.
3. The total runtime text is added at the top of the notebook for users to be aware of how long the notebook should take on default settings.

Associated back-end changes in https://github.com/cal-adapt/climakitae/pull/344/files.

### DIFF
--- a/work_in_progress/agnostic_tools.ipynb
+++ b/work_in_progress/agnostic_tools.ipynb
@@ -50,7 +50,8 @@
     "    plot_climate_response_LOCA,\n",
     "    show_available_vars,\n",
     "    get_available_units,\n",
-    ")"
+    ")\n",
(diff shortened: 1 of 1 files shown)


vicford (approved):
Made some improvements to clarity on the markdown. NB is good to go!

vicford (approved):

PR #109, opened 2024-05-28
Making small modifications to WL notebook to create more visually appealing plots

claalmve:
Plots were originally overlaying over each other. Plots with simulations overlaid with power stations is now fixed.

### DIFF
--- a/collaborative/IOU/vulnerability_assessment/vulnerability_assessment.ipynb
+++ b/collaborative/IOU/vulnerability_assessment/vulnerability_assessment.ipynb
@@ -29,11 +29,8 @@
    },
    "outputs": [],
    "source": [
-    "import climakitae as ck\n",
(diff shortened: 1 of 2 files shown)


vicford (changes requested):
Cell on hourly data is taking a long time to load... add some markdown that this will take approximately 12 minutes with the default selections.

Getting an issue at the plotting step about ccrs not being defined. Looks like this was removed and needs to be added back in. Once that is resolved, now getting an error that 'cfeature' is not defined.

Also, this plot is massive (likely to accomodate titles), let's reduce the size and make the colorbar not so large too.

Added a trello card about the colormap issue

vicford:
@claalmve in this PR we need to remove the vulnerability_assessment notebook from it so only the warming levels notebook gets modified!

PR #110, opened 2024-06-01
Pilot Phase 2 Notebook Changes

claalmve:
This notebook is the front-end for https://github.com/cal-adapt/climakitae/pull/357 and https://github.com/cal-adapt/climakitae/pull/360 . The appendix of this notebook will only run with with https://github.com/cal-adapt/climakitae/pull/360 since that is where the table generation code lies.


vicford (approved):
This is fine to go, and I added some new examples of how to do the "main 3 types of metrics" but should be waited until the climakitae PRs are finalized (at least #343)

PR #111, opened 2024-06-06
Making slight modifications to vulnerability notebook to accomodate for vul_table inclusion

claalmve:
This PR removes redundant imports and adds optionality for users to plug in specified thresholds into the `create_vul_table` function.


vicford (changes requested):
Just add the `create_vul_table` import back in please :) Otherwise it doesn't run!

vicford (approved):

vicford:
Slight recommendation is to include `df` at the end so it actually pulls up when running, but otherwise looks good!

PR #113, opened 2024-06-12
Making small modifications to WL notebook to create more visually appealing plots

claalmve:
Plots were originally overlaying over each other. Plots with simulations overlaid with power stations is now fixed.


vicford (approved):
Modified markdown:
- condense multiple markdown cells
- add runtime piece
Added colorbar shrink so it wasn't so massive + changed size of plot.

PR #124, opened 2024-08-20
Fix data_acquisition_for_hydro.ipynb

Tianchi-Liu:
- Update unit selected for precipitation from “mm/day” to “mm” and that for relative humidity from “%” to “[0 to 100]”. This is to comply with the `_check_valid_unit_selection` function introduced in climakitae after the notebook was developed.
- Update frequency attribute of daily data aggregated from hourly ones to "daily"
- Update ck to ckg so the functions moved from ck to ckg in the cksplit2a branches of both repos work

### DIFF
--- a/work_in_progress/data_acquisition_for_hydro.ipynb
+++ b/work_in_progress/data_acquisition_for_hydro.ipynb
@@ -261,7 +261,7 @@
    "source": [
     "data_to_view = max_airtemp_data.isel(time=np.arange(0,5)) # selecting only first 5 days\n",
     "data_to_view = ck.load(data_to_view)\n",
-    "ck.view(data_to_view)"
+    "ckg.view(data_to_view)"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


vicford (changes requested):
Thanks @Tianchi-Liu for updating this nb. The update of the units is perfect. Since the 4 new bias-adjusted models are now available at daily timesteps and not only hourly, we can now adjust the default selections cell (cell #3) to be "daily" and we won't need the resampling for min/max/mean on air temperature, relative humidity, or wind speed anymore. I've double checked on precipitation too (since that's actually two variables summed together) and it's available for the 4 newer models so that should be fine to switch over too!

vicford:
Just to clarify here @Tianchi-Liu -- `Precipitation (total)` is now a variable accessible in climakitae, so we don't have to retrieve the `cumulus portion only` and `grid-scale portion only` now; just `precipitation (total)`!

claalmve:
- changed hourly data retrieval and resampling into just retrieving daily data
- update variable names, units, and removed code as necessary for this change
- updated overall runtime of `bulk_run` and notebook runtime

vicford (changes requested):
@claalmve Thanks for tackling this!
- Will you update the bulk_run timing to be 5-10 minutes? My test was 8.5 minutes

Otherwise, looks great!

PR #125, opened 2024-08-21
Add new exploratory notebook on historical data comparisons

vicford:
Guidance notebook on understanding historical comparison between wx obs, reanalysis, and model output + recommendations


wkrantz:
Fixed one more typo and re-named the file to reflect the focus on historical data.

wkrantz (commented):
Oops, merged before reviewing, but looks good.

PR #126, opened 2024-08-22
Update internal_variability notebook

Tianchi-Liu:
Four WRF models have been added to the Analytics Engine since the development of the internal_variability notebook. This PR updates the notebook to include the new models.

The methodology stays the same. Changes are in the following areas:
- Code to add new models into the analyses
- Markdown to reflect the addition
- Miscellaneous minor improvements

Please test with [this climakitae PR](https://github.com/cal-adapt/climakitae/pull/456).


vicford (changes requested):
Hi @Tianchi-Liu -- thanks for tackling these changes for a nb that was done prior to the new batch of bias-adjusted models. 

Because the bias adjustment procedure only shifts the means of the boundary conditions for WRF it means that the native GCMs variability is maintained in the newer batch of models too. In short, we don't have to subset for only the non-BA models here anymore! I'd recommend:
-  removing the piece about subsetting the WRF ds, 
- adding the 4 new models to `cmip_names` = 'MIROC6', 'EC-Earth3', 'TaiESM', and 'MPI-ESM1-2-HR`
- we'll need to update some of the markdown too now to reflect that it is 8 models rather than 4 -- I can help you here

elehmer:
FYI `selections = ckg.Select()` should be replaced with `selections = DataParameters()` as `Select` is only useful if you need to show the GUI.

vicford (approved):
Thanks again @Tianchi-Liu. I've updated the data parameters selection piece in the other PR: https://github.com/cal-adapt/cae-notebooks/pull/136/commits/9372e40116d465d1e189348cf11d771ca83ce708

These are all good to be merged in

PR #129, opened 2024-09-19
Upgrade `thresholds_basics.ipynb` to reduce spatial extent and increase user visibility into runtimes

claalmve:
Notebook originally taking ~25 hrs because of spatial extent not being constrained (may have been unconstrained due to some merge conflict). Spatial extent is limited to California, to 45 km, with more visibility into runtimes to result in a total notebook runtime of ~50-55 min.


vicford (approved):
Finally ran, all good!

PR #134, opened 2024-10-16
Notebook updates to demonstrate warming level retrieval

nicolejkeeney:
1. Modify the code and markdown in the climakitae_direct_data_download.ipynb notebook to show how to retrieve data using a warming level approach (expected runtime is a few min-- shouldn't be longer) 
2. Modify the markdown in getting_started.ipynb notebook to explain new GUI options (no runtime change) 

This is part of a larger push to incorporate warming levels into climakitae and cae-notebooks. 

## Type of Change

- [ ] Bug fix
- [ ] New feature or notebook
- [ ] Breaking change
- [x] Documentation update

## Checklist
- [ ] Markdown reviewed and vetted by someone else
- [x] Error messages for areas that might expect common user error
- [ ] Verbose and non-verbose modes function
- [ ] It may not be required for notebooks that explicitly don’t need this functionality
- [ ] Clear expectations of outcomes learned (key takeaway messages) provided


climakitae_direct_data_download.ipynb
vicford, line 1: We should include some sort of markdown in this notebook referencing the default time bounds / WL settings if the user does not pass them to the `get_data` function. It occured to me that a user might be confused as to why 2.0degC was returned if they only passed warming levels but did not specify which level. 

This info can go into the "Example 2: Warming levels approach" markdown. Something like "The default selection for warming levels data retrieval is 2.0degC, but you can specify which level by using the `warming_level` parameter too. Options are: 1.5, 2.0, 2.5, 3.0, 4.0"

Same thing for time based at the "Example 1: Time-based approach" level.

nicolejkeeney: Fixed, thanks!

PR #173, opened 2025-07-14
Add DWR notebooks

neilSchroeder:
Adding notebooks from DWR for LOCA2 analysis.

Promised to incorporate years ago, finally happening.

This is the first user-contributed code to be merged to `cae-notebooks`. As such it needs a thorough discussion of the following:

1. How much altering will we make to notebooks before merging?
2. How much testing will we do prior to merging?
3. What kinds of claims are we willing/not willing to support about user-contributed code?

## Type of Change

- [ ] Bug fix
- [x] New feature or notebook
- [ ] Breaking change
- [ ] Documentation update
- [ ] None of the above

## Checklist
- [ ] The introduction of the notebook explains the purpose and expected outcome / use of the notebook


vicford (commented):
**Response to specific requested questions:**
_How much altering will we make to notebooks prior to merging?_
Preference here is none, beyond clearing the output of the notebook if there is any. The hope/goal/ideal is that user contributions would have a clear notebook name, and if notebooks/scripts need to be run in a specific order, that the user would document and provide that information. 

_How much testing will we do prior to merging?_
Theoretically, we should make sure that the notebooks do in fact complete a technically successful run, but with limited to no support on whether it was scientifically accurate. It would not be a good look for us if a bunch of broken notebooks were hosted, so whatever CYA level is acceptable. 

_What kinds of claims are we willing/not willing to support about user-contributed code?_
This is a tricky one and might be a larger conversation in terms of what our vetting or tolerance level is for which user contributions "pass muster" for us hosting. General first pass is: 
- Uses the 5th assessment data hosted by AE
- Has a CA / WECC / 5th assessment focus that can be of use to other users of 5th assessment data that is clearly documented with clear outcomes/objectives
- Documented "enough" that it is passable by our standards, while not being rigidly enforced
   - To this point, help/guidance/support should be clearly outlined by the user

**General but specific comments on the DWR notebooks:** 

> If you have questions about a specific contribution, please try to contact the original author when possible.

1. In the future, we should make sure this is explicitly provided by contributors and included in the README for each contribution set. 

2. Do we clear all output from all notebooks to be consistent with other AE notebooks? For each notebook to be explicitly clear as to what it is doing, user contributions should have a clear header (notebook). 
3. Do we have a protocol for when contributors include their own function definitions for functions that are also included in climakitae? e.g., `get_monthly_rolling_ave`. Perhaps we include a disclaimer in the `user-contributions` README about similar functions etc.
4. It appears that many of the notebooks in the `attic` folder are testing notebooks? Output is included, including errors, warnings, and duplicated functions. I am medium-ly inclined that we don't include this folder of notebooks, but point users towards the source `CADWR-Climate-Change-Program` github link if they are necessary. Also do we know what `attic` means? 
5. What's the `mask/mask.npy` file?

neilSchroeder:
@vicford 

> In the future, we should make sure this is explicitly provided by contributors and included in the README for each contribution set.
Agreed.

> Do we clear all output from all notebooks to be consistent with other AE notebooks? For each notebook to be explicitly clear as to what it is doing, user contributions should have a clear header (notebook).
I'm a little conflicted. I understand it makes a cleaner notebook and all, but I also don't want to force people to run them to see the output. Feels like a safety thing for me. Not a fan of running unknown code on my machine.

And for headers I totally agree, notebooks should be required to have a header.

> Do we have a protocol for when contributors include their own function definitions for functions that are also included in climakitae? e.g., get_monthly_rolling_ave. Perhaps we include a disclaimer in the user-contributions README about similar functions etc.
Catching/flagging this as something we have available would be challenging and time intensive. I'm inclined not to do anything other than be clear with language in the README that indicates these scripts aren't necessarily considered best practices.

> It appears that many of the notebooks in the attic folder are testing notebooks? Output is included, including errors, warnings, and duplicated functions. I am medium-ly inclined that we don't include this folder of notebooks, but point users towards the source CADWR-Climate-Change-Program github link if they are necessary. Also do we know what attic means?
I suspect the authors didn't want to hide these testing files, and just stuck them in the attic (like for a house) for storage. 

> What's the mask/mask.npy file?
If you go look at `attic/masking.ipynb` and `attic/masking-02.ipynb` it looks like the mask is just a binary mask for faster clipping. Probably done this way to avoid having to load the shapefile and clip and slice every time.

justineb17:
Thanks for pulling this together, Neil! I agree with Victoria's response to your first two questions, and leave the rest to the judgment of the development team. 

Based on the discussion, I do have some suggestions for the readme:

- Under the "Contributing" section, what if we had a point that said "If your code reproduces functionality already in climakitae, please note that in the comments or documentation." or something encouraging users to look at climakitae for functionality they may want to use in their code. 
- Under "Using these Contributions", maybe include "Note that included functions or approaches may differ from official AE or climakitae practices."

PR #178, opened 2025-08-04
Add climatology years to epw

acordonez:
Add years to the write_to_tmy() call (final cell) so that the climatology years are written to epw.

We want the correct climatology years written in the epw header.

## Type of Change

- [ ] Bug fix
- [ ] New feature or notebook
- [ ] Breaking change
- [ ] Documentation update
- [ ] None of the above

## Checklist
- [ ] The introduction of the notebook explains the purpose and expected outcome / use of the notebook

### DIFF
--- a/work-in-progress/typical_meteorological_year_methodology.ipynb
+++ b/work-in-progress/typical_meteorological_year_methodology.ipynb
@@ -1159,6 +1159,7 @@
     "    write_tmy_file(\n",
     "        filename,\n",
     "        tmy_data_to_export[sim],\n",
+    "        (1990,2020) # if climatology years are changed in the data, change the years here to match\n",
     "        stn_name,\n",
     "        stn_code,\n",
     "        stn_lat,\n",


vicford (approved):
Thanks @acordonez. I made some modifications to the markdown to be more explicit. Large change was that I modified all instances of the start_year and end_year to be based off of a single call, rather than having to modify it in a dozen places throughout the nb.

work-in-progress/typical_meteorological_year_methodology.ipynb
```diff
@@ -1201,7 +1181,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.8"
+   "version": "3.12.10"
```
neilSchroeder, line 1184: would love to move to an even newer version if we can, but not a big deal.

neilSchroeder (approved):
Additions/modifications look great!

If there's time I'd love to clean this notebook up with type hints and such, but otherwise it's good to go.

work-in-progress/typical_meteorological_year_methodology.ipynb
acordonez: Weird, my local version shows 3.12.10 when I open the notebook in an editor but that seems to not be getting tracked in git

PR #193, opened 2025-10-06
Link to HDP file in cadcat. Improve subset example

nicolejkeeney:
- The notebook links to the HDP station list file in cadcat, instead of climakitae, where it will be deleted. 
- I also updated the station subsetting example (see below for more details) 
- Added expected runtime to top of notebook

The previous subsetting section used code that the user was supposed to fill in, which results in errors being raised if the user actually ran the code cells without changing anything. It also referenced a function in `climakitae` for subsetting that is overengineered for the file in cadcat, which provides a geometry column. Lastly, it forced the shapefile to EPSG:3857, presumably to match the HDP data, but this is not actually the CRS of the HDP data. 

Anyways I used some creativity to come up with what I thought was a better example of how to subset using a shapefile! Now, the cells all run without error, and I think its a good and hopefully relevant example for our stakeholders. LMK what you think! 

## Type of Change

- [x] Bug fix
- [x] New feature or notebook
- [ ] Breaking change
- [x] Documentation update
- [ ] None of the above

## Checklist
- [x] Incorporates reference to any appropriate Guidance material
- [x] Notebook raises appropriate error messages for common user errors
- [x] [AE navigation guide](https://github.com/cal-adapt/cae-notebooks/blob/main/AE_navigation_guide.ipynb) updated (if relevant)


vicford (approved):
<3

PR #197, opened 2025-10-15
New GWL Approach Notebook

claalmve:
Update current `warming_levels_approach.ipynb`.

Need to update warming levels approach to become more up-to-date with new tools and better examples.

## Type of Change

- [ ] Bug fix
- [x] New feature or notebook
- [ ] Breaking change
- [ ] Documentation update
- [ ] None of the above

## Checklist
- [x] Incorporates reference to any appropriate Guidance material
- [x] Notebook raises appropriate error messages for common user errors

### DIFF
--- a/analysis/warming_level_methods.ipynb
+++ b/analysis/warming_level_methods.ipynb
@@ -116,7 +116,6 @@
     "\n",
     "The red shaded area in the plot above represents the IPCC's 90% confidence interval for the rate of future global warming under the SSP3-7.0 scenario, with the best estimate warming trajectory shown as the red dashed line.\n",
     "\n",
-    "\n",
     "![GWL timing](static_figures/gwl_figure2.png)"
    ]
   },


analysis/warming_level_approach_helpers.py
vicford: Is this supposed to be here? Or in climakitae?

analysis/warming_level_methods.ipynb
vicford: There are two intended application sections -- I think the first one is the complete one?

vicford (changes requested):
**Methods Notebook**
1. See note about static figures in nbs -- there's a way to directly put them in markdown which would be super helpful here. To that point: I'd also move the target year and gwl planning text up "the page". 
2. These figures would be nice in AE guidance somewhere too. 
3. Some links need to be "highlightitized" to make more obvious.
4. `It is apparent that many of the GCM simulations that underly the downscaled LOCA2 and WRF simulations used on the Analytics Engine are warming much faster than the IPCC's best estimates consider's realistic. ` is going to have to be explained further based on this figure. From a user learning about GWLs for the first time, this is not obvious in the figure as to what to look for. Annotations on the figure (like an arrow) could be very helpful here. This line comes to late `In the example shown in the figure above:  *The 2.0 °C GWL is estimated to be reached between 2037 and 2061 under SSP3-7.0, with the best estimate being 2047*, as indicated by the magenta lines and shaded area.`
5. On explaining historical vs. future climate in section 2, I'd clarify "using climate model data", since this has come up in terms of clarity at WGs.
6. I get why we've got 2 methods for retrieving GWL data in this notebook, but it feels clunky. Since new core is the wave of the future, my preference here is to have method 1 be in markdown code only as reference. 
7. I'd get rid of the second `ClimateData` call wrt precipitation, primarily because the notebook is long enough as it is. Especially since it's called again as example 2. 
8. The `lighter_r_rev` isn't used? If it's a colormap, we can add that into ck.
9. `**It's STRONGLY recommended to take the deltas of warming levels before aggregating across simulations**.` TY! And we should state _why_ , because guidance of recommended approaches!
10. Why is `tasmax_gwl_data` called -- could we not just use `gwl_data` instead since that was the first call? Light preference is to rename the first instance to `tasmax_gwl_data` for clarity; and then in the second example to call it `prec_gwl_data` (not `gwl_prec_data`) for consistency.
11. Are unit conversion not an option in `ClimateData` upon retrieval?
12. General note: This notebook has **a lot** of information. I would recommend a very obvious summary recap at the end of this notebook; and to preface this info at the beginning header, including a runtime timer. 
13. Don't forget to clear output. 

**new gwl approach nb**
I'm assuming that this is a prior draft version to the warming-levels-methods notebook on initial glance, so not going to review unless requested by @wkrantz

wkrantz:
@vicford Great catch, I changed back to the markdown syntax, can you double check that it looks okay in the hub? In VSCode it was rendering the figures as comically large, but it seems more well behaved in the hub for some reason.

PR #213, opened 2026-01-20
updating renewables access notebook

neilSchroeder:
Updating renewables access notebook with new core

Upcoming IOU meeting

## Type of Change

- [ ] Bug fix
- [ ] New feature or notebook
- [ ] Breaking change
- [ ] Documentation update
- [x] None of the above

## Checklist

### DIFF
--- a/data-access/renewables_data_access.ipynb
+++ b/data-access/renewables_data_access.ipynb
@@ -524,6 +524,7 @@
     "The files for hydropower generation can be found here.\n",
     "https://wfclimres.s3.amazonaws.com/index.html#era/hydropower_generation/\n",
     "\n",
+    " Hydropower generation projections were modeled using a different process than the solar and windpower data above. This data projects hydropower generation in MW at specific real-world facilities at rivers and reservoirs across the western United States. Generation is based on a variety of climate variables across the watersheds that feed into each generation facility. \n",
     "The following cells show how to load this dataset:"
    ]
   },


vicford (changes requested):
**Technical review:**
- Everything all runs!
- Should a user know what the different options for `verbosity` mean? We've now seen `-1` and `-2` but *I* specifically don't know what those mean.

**Markdown changes (lots of nit-picky things):**
- General comment: I'd like the notebook header to be more descriptive about the renewables data, beyond the data guide (which is amazing). What is the purpose of the renewables data? What might be included? The "access cf and generation data" sentence is a step in this direction. See the [HDP station access notebook](https://github.com/cal-adapt/cae-notebooks/blob/main/data-access/weather_station_data_access.ipynb) for a lil more idea in this space. 
   - Note: this might require coordination with @bethem or if there is a general overview paragraph on the renewables data that can be included here. 
   - Ah -- seeing a lot of info at the bottom of the notebook that I wanted at the top. 
- `Installation Types Available` looks like the two windpower bullet points aren't in the full list format (they're tacked on to the pv ones). This also has happened in the `Working with Processors` section for `export` bullet point
- Looks like the `Example 1: Single Installation Type and Variable` header got folded into the `Required Parameters for Renewables Queries` section
- Looks like the "not specifying a scenario" returns the exact same thing as the previous cell. Is this intended? This could also be just a sentence added to the previous cell rather than an explicit call
-  Plotting -- is there a reason we are not using ckg functionality? or `ckg.view`? 
- Retrieving data for SF, is this the SF airport? If so, we should specify. Same with multiple cities
- In general with figures, I'd like us to start including some light markdown about how a user should interpret a figure / what they should take away from it (improving climate literacy). The first figure: I'd add something to the effect of `This is a timeseries of modeled PV distributed generation for San Francisco in 1981. The "dips" you see are when PV generation was reduced or diminished, which is indicative of cloudy conditions or dunkelflautes`. For the second figure: A user might naturally ask "LA and SD are really close geographically, why does SD have higher CF than LA?" 
- I personally don't like the call for issues / feedback within the notebook -- but I am fine to be overruled, as long as we make this consistent with all other notebooks. My light preference is, if we want to keep this, put it in the AE nav guide _or_ have a "go to" sentence with link that is put into every notebook header (could be a part of the nb refactor).

PR #218, opened 2026-02-11
Feature/custom metrics example

neilSchroeder:
This notebook demonstrates how to use `climakitae` to define custom metrics, and uses cooling days to do so.

Adds important documentation / examples, updated example of cooling days with new core.

## Type of Review
Need a scientific review, as well as a smaller technical review.

## Type of Change

- [ ] Bug fix
- [x] New feature or notebook
- [ ] Breaking change
- [ ] Documentation update
- [ ] None of the above

## Checklist


analysis/derived_variables_demo.ipynb
```diff
@@ -0,0 +1,798 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "df7c9f7b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Average over spatial dimensions for county-wide average\n",
+    "from dask.diagnostics import ProgressBar\n",
```
nicolejkeeney: same here, this is imported at top of notebook and should be removed here

nicolejkeeney (approved):

analysis/derived_variables_demo.ipynb
```diff
@@ -0,0 +1,838 @@
...
+   "cell_type": "markdown",
+   "id": "e42bba08",
+   "metadata": {},
+   "source": [
+    "## Why Derived Variables? Old vs New Approach\n",
+    "\n",
+    "### ❌ Previous Approach (Legacy Interface)\n",
+    "\n",
+    "```mermaid\n",
+    "flowchart TD\n",
```
wkrantz: For what it's worth, these cool flowcharts render for me on the juputerhub, but not in vscode.

analysis/derived_variables_demo.ipynb
```diff
@@ -0,0 +1,838 @@
...
+    "### Code Comparison\n",
+    "\n",
+    "**Old Approach:**\n",
+    "```python\n",
+    "# Step 1 & 2: Multiple data fetches\n",
+    "tasmax_data = get_data({'variable': 'tasmax', 'area': 'Los Angeles', ...})\n",
+    "tasmin_data = get_data({'variable': 'tasmin', 'area': 'Los Angeles', ...})\n",
+    "\n",
+    "# Step 3 & 4: Manual calculation\n",
+    "def calc_cdd(tasmax, tasmin):\n",
```
wkrantz: There is already a `compute_hdd_cdd` function defined in climakitae.tools.derived_variables. Yours is a simpler function, but for internal consistency it would be nice to use what we already have defined.

analysis/derived_variables_demo.ipynb
```diff
@@ -0,0 +1,560 @@
...
+    "    style A fill:#ccffcc,stroke:#00aa00,stroke-width:2px\n",
+    "    style B fill:#ccffcc,stroke:#00aa00,stroke-width:2px\n",
+    "    style C fill:#eeffee,stroke:#00aa00,stroke-width:3px\n",
+    "```\n",
+    "\n",
+    "**Benefits:**\n",
+    "- One reusable metric definition\n",
+    "- Single `.get()` query with clipping + warming-level processing\n",
+    "- Lazy xarray/dask workflow (compute only when needed)\n",
+    "- Clear translation to business-facing “comfortable day” framing"
```
wkrantz:
```suggestion
 "## Workflow comparison\n",
    "\n",
    "#### Without register_user_function\n",
    "Without the ```register_user_function``` tool, the workflow for calculating a derived metric that uses temperature and relative humidity would look like this:\n",
    "\n",
    "```mermaid\n",
    "flowchart TD\n",
    "    A[Step 1: Define function for custom metric] --> B[Step 2: Individually load each of the required input metrics e.g. t2, rh]\n",
    "    B --> C[Step 3: Compute metric by sending input metrics through custom function]\n",
    "    C --> D[Step 4: Perform spatial and temporal aggregation]\n",
    "    D --> E[Step 5: Load resulting data into local memory for plotting or analysis]\n",
    "\n",
    "    style A fill:#FFD580,stroke:#FF8C00,stroke-width:2px\n",
    "    style B fill:#FFD580,stroke:#FF8C00,stroke-width:2px\n",
    "    style C fill:#FFD580,stroke:#FF8C00,stroke-width:2px\n",
    "    style D fill:#FFD580,stroke:#FF8C00,stroke-width:2px\n",
    "    style E fill:#FFD580,stroke:#FF8C00,stroke-width:2px\n",
    "```\n",
    "\n",
    "\n",
    "    \n",
    "### With register_user_function\n",
    "With the ```register_user_function``` tool, only one call to fetch data from the catalog is necessary, which returns the custom metric directly:\n",
    "\n",
    "```mermaid\n",
    "flowchart TD\n",
    "    A[Step 1: Define function for custom metric] --> B[Step 2: Register function with climakitae]\n",
    "    B --> C[Step 3: Use one call to ClimateData to load custom derived metric]\n",
    "    C --> D[Step 4: Perform spatial and temporal aggregation]\n",
    "    D --> E[Step 5: Load resulting data into local memory for plotting or analysis]\n",
    "\n",
    "    style A fill:#ccffcc,stroke:#00aa00,stroke-width:2px\n",
    "    style B fill:#ccffcc,stroke:#000000,stroke-width:5px\n",
    "    style D fill:#ccffcc,stroke:#00aa00,stroke-width:2px\n",
    "    style C fill:#ccffcc,stroke:#000000,stroke-width:5px\n",
    "    style E fill:#ccffcc,stroke:#00aa00,stroke-width:3px\n",
    "```\n",
    "\n",
    "\n",
    "```mermaid\n",
    "flowchart TD\n",
    "    A[Step 1: Register custom derived variable\\nGoldilocks from t2 and rh] --> B[Step 2: Single query\\nClimateData.variable.processes.get]\n",
    "    B --> C[Step 3: Analyze outputs\\nSpatial + temporal views by warming level]\n",
    "\n",
    "    style A fill:#ccffcc,stroke:#00aa00,stroke-width:2px\n",
    "    style B fill:#ccffcc,stroke:#00aa00,stroke-width:2px\n",
    "    style C fill:#eeffee,stroke:#00aa00,stroke-width:3px\n",
    "```\n",
    "\n",
    "## Benefits to registering custom metric\n",
    "\n",
    "This second workflow enables a **simpler** workflow after the initial metric has been defined, with only one `.get()` query that can be combined clipping + warming-level processing that would otherwise need to be repeated for each input variable.\n",
    "\n",
    "Registering the metric function should produce a significantly **faster** workflow overall, because it enables full utilization of a lazy xarray/dask workflow (compute only when needed) and backened efficiencies in the ClimateData framework.\n",
    "\n",
    "Once defined and registered, a custom metric can also **easily be re-used** in other contexts and analysis workflows.\n"
    
      "## Example: Custom *Goldilocks Day* metric\n",
    "\n",
    "To demonstrate the workflow for defining a custom derived metric, this notebook introduces a **Goldilocks Day** metric that uses temperature and relative humidity as inputs. We will use this metric to examine shifts to the climate in Los Angeles County.\n",
    "\n",
    "A Goldilocks day is defined here as a day when:\n",
    "- Temperature at 2m (`t2`) is between **65°F and 80°F**\n",
    "- Relative humidity (`rh`) is between **40% and 50%**\n",
    "\n",
    "\n"
```

wkrantz (commented):
Thanks for the updates to this notebook! My high level comment is I think this notebook would best serve users by focusing on demonstrating the functionality of registering custom metrics, and demonstrating how it is different from a "traditional" approach of computing them from individual metrics. The specifics of the example metric used and the results it produces are interesting, but not that important for this notebook. I re-wrote some of the markdown at the top of the notebook to lean in that direction, but feel free to edit what I wrote to better capture the advantages of this method that you want to highlight, or if i misrepresented the difference in workflows.

In the bottom half of the notebook, here is what I would suggest:
- The figures that utilize the values of the t2 and rh are nice (I love the last one with the heatmap of values!), but are not needed to illustrate the custom metric workflow, and are a little confusing because to create them you have to load the individual metrics you just went out of your way to avoid loading. I would recommend sticking to the map, and then maybe a single violin plot of the Goldilocks days directly (rather than violin plots for the constituent variables.)

- After some nice plots demonstrating the Goldilocks data, include a block of code that represents the "traditional" workflow for accomplishing the same thing. It could be commented out initially, but included with an indication of the estimated difference in time to run that workflow compared to registering the metric. For me, it was about 4 minutes to run it with the custom function registered, and 11 minutes to run it the old fashioned way (that's locally on my laptop).

- Thinking about how to illustrate the benefits of this approach, you could include an example of a second call to ClimateData pulling the custom metric with a different set of "processes" (maybe time-based, in a different domain). 

e.g. for the "individual metric" workflow
```
def fetch_driver(variable_name):
    return (
        cd_driver
        .catalog("cadcat")
        .variable(variable_name)
        .activity_id("WRF")
        .institution_id("UCLA")
        .table_id("day")
        .grid_label("d03")
        .processes({
            "warming_level": {"warming_levels": [1.2, 2.0, 3.0]},
            "clip": "Los Angeles County",
        })
        .get()
    )

t2_data = fetch_driver("t2")
rh_data = fetch_driver("rh")

ds = xr.merge([t2_data, rh_data])

ds_gl = calc_goldilocks_day_wrf(ds)

bad_sim = "WRF_UCLA_EC-Earth3-Veg_ssp370_day_d03_r1i1p1f1"
if "sim" in ds_gl.dims and bad_sim in ds_gl.sim.values:
    ds_gl = ds_gl.drop_sel(sim=bad_sim)

ds_gl = goldilocks_data = add_dummy_time_to_wl(ds_gl)

# Count Goldilocks days per year
goldilocks_days_per_year = ds_gl["goldilocks_day_wrf"].resample(time="Y").sum()

# Restore map-style NaN handling: show only cells with at least one Goldilocks day
goldilocks_days_per_year = xr.where(
    goldilocks_days_per_year > 0,
    goldilocks_days_per_year,
    float("nan"),
 )

# Average across years for spatial plotting
spt = goldilocks_days_per_year.mean(dim="time")

# Compute median across simulations for each warming level (ignore NaNs at boundary)
with ProgressBar():
    spt = spt.median(dim="sim", skipna=True).compute()
```

PR #244, opened 2026-04-07
Consolidate threshold tools notebooks to one notebook that uses `new_core`

claalmve:
## Summary of changes
Creating a new threshold tools notebook that consolidates the 3 existing threshold tools notebook, and uses new core instead of existing AE functionality.

## Link to corresponding Jira ticket(s)
[Jira link](https://eaglerockanalytics.atlassian.net/browse/AE-1369?atlOrigin=eyJpIjoiZGFkYWM1ZTljYWJjNGEyMWI3OWIwZTIwNTE1MGM5OTIiLCJwIjoiaiJ9)

## Naming & Organization
- [x] Notebook name follows conventions:
  - [ ] Avoids acronyms and jargon where possible
  - [ ] References primary use of the notebook
  - [ ] Uses Action + Objective format (e.g., "calculate_heat_index", "calculate_annual_trends")
  - [ ] Prioritizes clarity over brevity (within reason)
- [x] Notebook placed in appropriate directory for maximum usability
- [x] All referenced guides/documentation updated:
  - [x] README updated
  - [x] Navigation Guide updated

## Content & Documentation
- [x] Introduction clearly explains the purpose of the analysis
- [x] Intended application of the notebook listed
  - User-focused question/example provided showing how a user may want to use the notebook
- [x] Incorporates references to appropriate Guidance materials
- [x] Error messages included for areas with common user errors

## Runtime Information
- [x] Overall runtime on AE JupyterHub documented

## Output Management
- [x] Output from cells removed before committing

## Code Quality
- [x] Notebook cleaned up with:
  - [x] black
  - [ ] isort


analysis/new_thresholds_nb.ipynb
```diff
@@ -0,0 +1,361 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, ax = plt.subplots()\n",
+    "one_in_x_temp[\"return_periods\"].median(\"sim\").sel(one_in_x=105).plot(\n",
+    "    vmax=40, cmap=\"YlOrRd_r\", x=\"lon\", y=\"lat\",\n",
```
acordonez: Would any of the built-in AE colormaps work better here? Would love to find a colormap that's less of a blob for these results, if possible.

analysis/new_thresholds_nb.ipynb
```diff
@@ -0,0 +1,361 @@
...
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 3. Different return periods across WLs for a given threshold\n",
+    "\n",
+    "Another question we can answer using `climakitae`, is **how does the return period of a specific event change across warming levels?** \n",
```
acordonez: Posted about this slack, but I think it'd be more correct to get the return period using the 1-in-10 value for each simulation, rather than the median. Open to changing the example or using the threshold tools functions to show how to do this manually.

analysis/thresholds_tools.ipynb
```diff
@@ -0,0 +1,377 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Threshold tools\n",
```
acordonez: Once the thresholds guidance is online we'll want to link it, but that might be for a later PR.

analysis/new_thresholds_nb.ipynb
```diff
@@ -0,0 +1,377 @@
...
+    "ax.set_xlabel(\"Longitude\")\n",
+    "ax.set_ylabel(\"Latitude\")\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## 3. Different return periods across WLs for a given threshold\n",
```
acordonez: As already discussed, update this section to use a threshold selected from a newsworthy event rather than the simulation median.

analysis/threshold_tools.ipynb
```diff
@@ -0,0 +1,377 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 1-in-X options\n",
+    "\n",
+    "The examples in this notebook use the `metric_calc` processor available through the `climakitae` `ClimateData` object to run 1-in-X evaluations. The following arguments can be passed to `one_in_x` in the `metric_calc` processor:\n",
+    "\n",
+    "| Argument | Options | Required? | Notes |\n",
```
acordonez, line 54: Nice table!

analysis/threshold_tools.ipynb
```diff
@@ -0,0 +1,377 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Initialize the interface\n",
+    "cd = ck.ClimateData(verbosity=0)\n",
+    "\n",
+    "one_in_x_data = (\n",
```
acordonez, line 101: Based on what I'm hearing about users wanting notebooks they can easily use in workflows, I'm wondering if moving some of the variables like return_periods up to the top of the cell (outside the ClimateData call) to set/reuse would help. Not sure if this suggestion would actually accomplish that, curious what others think.

analysis/threshold_tools.ipynb
```diff
@@ -0,0 +1,377 @@
...
+    "<br>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# 2. Return periods: temperature over a region\n",
+    "\n",
+    "In this example, we used a time-based approach to find the return period for a specific temperature threshold (115°F for a 3-day heat wave) across San Bernardino County for simulation years 2040-2060.  We then display the results as a map. \n",
```
acordonez, line 179: I increased the threshold amount to make the map less of a single giant blob - that's a matter of personal opinion though.

analysis/threshold_tools.ipynb
```diff
@@ -0,0 +1,377 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, ax = plt.subplots()\n",
+    "one_in_x_temp[\"return_periods\"].median(\"sim\").sel(one_in_x=tmax_3day).plot(\n",
```
acordonez, line 257: Is there a way to get the county outline from climakitae and add it to the map?

acordonez (changes requested):
Thanks for pulling this together! I left some food-for-thought type comments but the only major change is that Section 3 example.

analysis/threshold_tools.ipynb
claalmve: Good idea! I can move these parameters outside of the calls.

analysis/new_thresholds_nb.ipynb
claalmve: Per Slack, making this an example where we look at historical 99th percentile for each GCM, and use that as the return value for determining return periods for 0.8, 2.0, and 3.0 WLs.

analysis/thresholds_tools.ipynb
```diff
@@ -0,0 +1,413 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Retrieving the San Bernardino County boundary for clipping\n",
+    "from climakitae.new_core.data_access import DataCatalog\n",
+    "sb_county = DataCatalog().boundaries._ca_counties.query(\"NAME == 'San Bernardino County'\")"
```
neilSchroeder: This is fine for now but @claalmve can you please add a ticket to the backlog to set up accessor functions for the boundaries? this kind of access isn't particularly maintainable.

analysis/thresholds_tools.ipynb
```diff
@@ -0,0 +1,413 @@
...
+    "        .variable(\"prec\")\n",
+    "        .processes({\n",
+    "            \"warming_level\": {\"warming_levels\": [0.8, 2.0, 3.0]},\n",
+    "            \"convert_units\": \"inches/d\",\n",
+    "            \"clip\": \"KSAC\",\n",
+    "            \"metric_calc\": {\n",
+    "                \"one_in_x\": {\n",
+    "                    \"return_values\": [threshold],\n",
+    "                    \"extremes_type\": \"max\",\n",
+    "                    \"variable_preprocessing\": {\n",
```
neilSchroeder: This option isn't listed in the options table at the start of the notebook. Feels odd to tell users the list of options and then use one they weren't told about in the example.

analysis/threshold_tools.ipynb
```diff
@@ -0,0 +1,413 @@
...
+    "### 1-in-X options\n",
+    "\n",
+    "The examples in this notebook use the `metric_calc` processor available through the `climakitae` `ClimateData` object to run 1-in-X evaluations. The following arguments can be passed to `one_in_x` in the `metric_calc` processor:\n",
+    "\n",
+    "| Argument | Options | Required? | Notes |\n",
+    "|---|---|---|---|\n",
+    "| `return_periods` | List of ints | Required (or `return_values`) | Return periods in years, e.g. `[10, 25, 50]` |\n",
+    "| `return_values` | List of floats | Required (or `return_periods`) | Threshold values, e.g. `[105]` |\n",
+    "| `extremes_type` | `\"max\"`, `\"min\"` | Optional | Default: `\"max\"` |\n",
+    "| `distribution` | `\"gev\"`, `\"genpareto\"`, `\"gumbel\"`, `\"weibull\"`, `\"pearson3\"`, `\"gamma\"` | Optional | Default: `\"gev\"` |\n",
```
neilSchroeder, line 59: Might be useful to either link to some guidance or something about when to use each of these. The old notebook had a short blurb about it that I think it would be nice to preserve, or at least point users in the right direction.

analysis/threshold_tools.ipynb
```diff
@@ -0,0 +1,413 @@
...
+    "    .table_id(\"day\")\n",
+    "    .grid_label(\"d02\")\n",
+    "    .variable(\"t2max\")\n",
+    "    .processes(\n",
+    "        {\n",
+    "            \"time_slice\": (2040, 2060),\n",
+    "            \"convert_units\": \"degF\",\n",
+    "            \"clip\": \"San Bernardino County\",\n",
+    "            \"metric_calc\": {\n",
+    "                \"one_in_x\": {\n",
```
neilSchroeder, line 210: I think we never hooked it up, but is it worth hooking up the ESS code for this example?

analysis/thresholds_tools.ipynb
```diff
@@ -0,0 +1,413 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Initialize the interface\n",
+    "cd = ck.ClimateData(verbosity=0)\n",
```
neilSchroeder: This might be too loud for most users, and we may want to set it to be a littler quieter

analysis/thresholds_tools.ipynb
```diff
@@ -0,0 +1,413 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Plotting the median return periods for a 3-day, 115°F heat wave across the county\n",
+    "fig, ax = plt.subplots()\n",
+    "one_in_x_temp[\"return_periods\"].median(\"sim\").sel(one_in_x=tmax_3day).plot(\n",
+    "    vmax=40,\n",
```
neilSchroeder: can you add a note here that these only apply for San Bernadino county?

neilSchroeder (changes requested):
Some things to consider for QoL and readability / usability

analysis/threshold_tools.ipynb
acordonez: We can suggest using the GEV as a default but once you get into the specifics of when to use which distribution it becomes a lot more than a short blurb. When the guidance I'm working on for the website is live we can definitely link that, too.

acordonez: The old notebook blurb wasn't completely accurate, and I tried overhauling it in the guidance but that quickly spiraled into a big task.

neilSchroeder: Makes sense, just wanted to float it for consideration and it sounds like it's already been thought about 👍🏻

analysis/thresholds_tools.ipynb
acordonez: Added

analysis/thresholds_tools.ipynb
acordonez: Changed to -1

analysis/thresholds_tools.ipynb
acordonez: Added

analysis/threshold_tools.ipynb
acordonez: Added a list of the full names of the distributions

analysis/threshold_tools.ipynb
acordonez: ESS PR for climakitae is still in the works but we can add a comment to the notebook about it. Example 1 has a large enough ESS, the other examples might, too.

vicford (commented):
I've updated the markdown in two ways:
- Improving the order and consistency of section breaks. 
- Clarity around the K-S test p-value fit. The "normal" use of p-values is used to determine statistical significantly different; here we're using the inverse to determine that the distribution fit does match the data (not statistically significantly different). Some folks would see p-values of 0.7 and be like... wait what. 

**Questions**:
- Example 1: I was a tad surprised to see that the return values for precipitation 1-in-Xs decreased from 0.8 GWL to 2.0 GWL. Any thoughts why? The increase in value from return period within a GWL makes sense; the comparison across highlights the problem.
- Example 3: Is the historical 99.5th percentile being calculated across the entire historical period per model? Aka 1981-2014? And then we grab 0.8, 2.0, and 3.0 GWLs and calculate return-values using the full historical time period range? If so that's a mix of GWLs to time-based reference. Considering we use 0.8 as a historical reference period it was surprising. I'd recommend we use 0.8 GWL as the historical 99.5 "calculator" and then do the future comparison using 1.5, 2.0, and 3.0 GWLs
   - In the "Conclusion" interpretation, it would be nice to include what the return value (median?) is here and expand this out a bit further but I want to double check that I'm interpreting this correctly.
- Is there a missing value in the `variable_preprocessing` notes for the Default?

claalmve:
> * Example 1: I was a tad surprised to see that the return values for precipitation 1-in-Xs decreased from 0.8 GWL to 2.0 GWL. Any thoughts why? The increase in value from return period within a GWL makes sense; the comparison across highlights the problem.

It looks like EC-Earth3 particularly gets lower return values at WL 2.0 vs 0.8, but not sure why that's happening.
image


Changing 2.0 to 3.0 for this example in the notebook in order to show the trend we're trying to illustrate:
image


> * Example 3: Is the historical 99.5th percentile being calculated across the entire historical period per model? Aka 1981-2014? And then we grab 0.8, 2.0, and 3.0 GWLs and calculate return-values using the full historical time period range? If so that's a mix of GWLs to time-based reference. Considering we use 0.8 as a historical reference period it was surprising. I'd recommend we use 0.8 GWL as the historical 99.5 "calculator" and then do the future comparison using 1.5, 2.0, and 3.0 GWLs

Good catch! Changing to 0.8 WL as the historical.

>   * In the "Conclusion" interpretation, it would be nice to include what the return value (median?) is here and expand this out a bit further but I want to double check that I'm interpreting this correctly.

Return values included and conclusion expanded a bit!

**Conclusion:** As warming levels increase from 0.8°C to 2.0°C, the median return period across GCMs decreases from **2.39 years** to **2.03 years**. This means that an historically extreme precipitation event becomes even more frequent under higher levels of global warming.

> * Is there a missing value in the `variable_preprocessing` notes for the Default?

No, it defaults to `{}`. Changing text to `code` text so that it shows up less like a typo.

analysis/threshold_tools.ipynb
```diff
@@ -0,0 +1,445 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Using the same set of GCMs as `sims` in example 1\n",
+    "GCMS = [\"EC-Earth3\", \"MPI-ESM1-2-HR\", \"TaiESM1\", \"MIROC6\"]\n",
+    "\n",
+    "results = []\n",
+    "\n",
+    "# Data retrieval and processing loop for each GCM\n",
+    "for gcm in GCMS:\n",
```
neilSchroeder, line 299: This loop isn't very efficient, but `climakitae` doesn't really support multi-gcm calling. I need to think about whether this is something we should change or not. No action needed here, just "thinking out loud".

neilSchroeder (approved):
This looks technically sound, nicely done!

vicford (approved):

nicolejkeeney (changes requested):
We need to add info about the three archived notebooks to the archive README: https://github.com/cal-adapt/cae-notebooks/blob/main/archive/README.md

Also, just a thought— are there any tests in `climakitae` for unsupported/unused thresholds code that should be removed?

analysis/threshold_tools.ipynb
```diff
@@ -0,0 +1,445 @@
...
+    "    \"NAME == 'San Bernardino County'\"\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
nicolejkeeney, line 254: Getting a TypeError in this cell: 
```python 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[8], line 3
      1 # Plotting the median return periods for a 3-day, 115°F heat wave across the county
      2 fig, ax = plt.subplots()
----> 3 one_in_x_temp["return_periods"].median("sim").sel(one_in_x=tmax_3day).plot(
      4     vmax=40, # max color limit specific to the San Bernardino results
      5     vmin=1,
      6     cmap=read_ae_colormap("ae_orange").reversed(),
      7     x="lon",
      8     y="lat",
      9     ax=ax,
     10     cbar_kwargs={"label": "Return Period (years)"},
     11 )
     12 # Overlaying the county boundary
     13 sb_county.boundary.plot(ax=ax, color="darkgrey", linewidth=1.0)

TypeError: 'NoneType' object is not subscriptable
```

acordonez:
> Also, just a thought— are there any tests in `climakitae` for unsupported/unused thresholds code that should be removed?

We haven't deprecated any of the existing thresholds code in climakitae

acordonez:
@nicolejkeeney I updated all the READMEs in the affected directories along with the AE navigation guide.

analysis/threshold_tools.ipynb
acordonez: @nicolejkeeney The cell before this one (to get `one_in_x_temp`) ran successfully?

nicolejkeeney: The cell above outputs this error: 

```python 
2026-05-04 23:17:30 - climakitae.new_core.user_interface - ERROR - Error during dataset creation: return_periods is required for 1-in-X calculations
```

acordonez: Any chance you don't have the latest changes to climakitae in your environment? This depends on https://github.com/cal-adapt/climakitae/pull/750, which is merged but not yet in a released version.

nicolejkeeney: Dang, great catch. You’re right, sorry about that.

nicolejkeeney (approved):
Looks great! Really nicely done, I appreciate all the hard work @claalmve and @acordonez !!

